PR #27, opened 2020-09-16
Change is_streaming:bool to stream_samples:u8

mmoskal:
@pelikhan if I change this in JM-modules, can you update jacdac-ts to re-send `stream_samples := 255` for everything that needs streaming every `bus.on(SELF_ANNOUNCE, () => { ... })` or so? It should also work with current modules.


pelikhan:
Will do!

jamesadevine (approved):

PR #44, opened 2020-10-27
messagebus spec

pelikhan:
Allows CODAL-like messages over jacdac.

### DIFF
--- a/services/messagebus.md
+++ b/services/messagebus.md
@@ -1,0 +1,23 @@
+# Message Bus
+
+    identifier: 0x17feef9b
+    
+A lightweight event bus that allows to broadcast
+short messages.
+
+## Commands
+
+    command send @ 1 {
+        id: u16;
+        name: u16;
+    }
+
+## Events
+
+    event message @ 2 {
+        id: u16;
+        name: u16;
+    }
+
+A message event was raised on the message bus 
+and is broadcasted as an event.
--- a/services/messagebus.md
+++ b/services/messagebus.md
@@ -3,20 +3,18 @@
     identifier: 0x17feef9b
     
 A lightweight event bus that allows to broadcast
-short messages.
+short messages, similar to the Scratch message APIs.
 
 ## Commands
 
-    command send @ 1 {
-        id: u16;
-        name: u16;
+    command send @ 0x80 {
+        msg: u32
     }
 
 ## Events
 
-    event message @ 2 {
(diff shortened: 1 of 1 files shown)


services/messagebus.md
```diff
@@ -0,0 +1,23 @@
+# Message Bus
+
+    identifier: 0x17feef9b
+    
+A lightweight event bus that allows to broadcast
+short messages.
+
+## Commands
+
+    command send @ 1 {
```
mmoskal, line 10: this probably needs to be @ 0x80 - you should be getting an error here?

services/messagebus.md
```diff
@@ -0,0 +1,23 @@
...
+        id: u16;
```
mmoskal, line 11: I'm not sure we accept semicolons, we surely don't require them

mmoskal: also, probably best to use `u32` for future extensibility

services/messagebus.md
```diff
@@ -0,0 +1,23 @@
...
+        name: u16;
+    }
+
+## Events
+
+    event message @ 2 {
```
mmoskal, line 17: events have a separate numbers from commands, so you can start at `1`

services/messagebus.md
```diff
@@ -0,0 +1,23 @@
+# Message Bus
+
+    identifier: 0x17feef9b
+    
+A lightweight event bus that allows to broadcast
+short messages.
```
mmoskal, line 6: Would be nice to have some usage scenario. I'm actually not sure about it.

services/messagebus.md
pelikhan: will check

services/messagebus.md
```diff
@@ -3,20 +3,18 @@
     identifier: 0x17feef9b
     
 A lightweight event bus that allows to broadcast
-short messages.
+short messages, similar to the Scratch message APIs.
```
mmoskal, line 6: Is the service just going to raise the event after it gets a send?

pelikhan: so basically the idea is that when the device receives a JACDAC event; it raises it on the CODAL event bus side.

mmoskal: so there is no generally-agreed-upon meaning to these events? If it's for CODAL, it should be called CODAL Message Bus; but I think we also need a clear usage scenario.

PR #45, opened 2020-10-28
mouse spec

pelikhan:
Support for HID Mouse drivers.

### DIFF
--- a/services/mouse.md
+++ b/services/mouse.md
@@ -1,6 +1,7 @@
 # Mouse
 
     identifier: 0x1885dc1c
+    status: experimental
 
 Control a device that acts as a mouse.
 
--- a/services/mouse.md
+++ b/services/mouse.md
@@ -6,16 +6,16 @@ Control a device that acts as a mouse.
 
 ## Commands
 
-    enum Button : u8 {
-        right = 0x01
-        middle = 0x04
-        left = 0x02
+    flags Button : u8 {
+        Right = 0x01
+        Middle = 0x04
+        Left = 0x02
     }
     enum ButtonEvent : u8 {
-        up = 0x01
-        down = 0x02
-        click = 0x03
-        double_click = 0x04
+        Up = 0x01
+        Down = 0x02
+        Click = 0x03
+        DoubleClick = 0x04
     }
     command set @ 0x80 {
         buttons: Button
@@ -25,13 +25,18 @@ Control a device that acts as a mouse.
 Sets the up/down state of one or more buttons.
 
     command move @ 0x81 {
-      x: i8
-      y: i8
+      dx: i16
+      dy: i16
+      time: u16
     }
 
-
 Moves the mouse by the distance specified.
+If the time is positive, it specifies how long to make the move.
 
-    command wheel: i8 @ 0x82
+    command wheel @ 0x82 {
+        dy: i16
+        time: u16
+    }
 
-Turns the wheel up or down.
+Turns the wheel up or down. Positive if scrolling up.
+If the time is positive, it specifies how long to make the move.
--- a/services/mouse.md
+++ b/services/mouse.md
@@ -25,17 +25,17 @@ Control a device that acts as a mouse.
 Sets the up/down state of one or more buttons.
 
     command move @ 0x81 {
-      dx: i16
-      dy: i16
-      time: u16
+      dx: i16 #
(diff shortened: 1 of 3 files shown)


services/mouse.md
```diff
@@ -0,0 +1,37 @@
+# Mouse
+
+    identifier: 0x1885dc1c
+
+Control a device that acts as a mouse.
+
+## Commands
+
+    enum Button : u8 {
```
mmoskal, line 9:
```suggestion
    flags Button : u8 {
```

services/mouse.md
```diff
@@ -0,0 +1,37 @@
...
+        right = 0x01
+        middle = 0x04
+        left = 0x02
+    }
+    enum ButtonEvent : u8 {
+        up = 0x01
+        down = 0x02
+        click = 0x03
+        double_click = 0x04
```
mmoskal, line 18: I think we use `DoubleClick` casing elsewhere

services/mouse.md
```diff
@@ -0,0 +1,37 @@
...
+    }
+    command set @ 0x80 {
+        buttons: Button
+        event: ButtonEvent
+    }
+
+Sets the up/down state of one or more buttons.
+
+    command move @ 0x81 {
+      x: i8
```
mmoskal, line 28: dx/dy?

services/mouse.md
```diff
@@ -0,0 +1,37 @@
...
+      y: i8
```
mmoskal, line 29: what is the unit of the distance? (could be in comment)
also maybe use i16

services/mouse.md
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+
+Moves the mouse by the distance specified.
+
+    command wheel: i8 @ 0x82
+
+Turns the wheel up or down.
```
mmoskal, line 37: would be good to say "positive is scroll up" (if mouse set to "natural scrolling") or something

services/mouse.md
mmoskal: also should there be a time parameter? to spread the movement over time?

services/mouse.md
mmoskal: I don't think USB mouse is generating click or double click events, is it?

pelikhan: I am thinking that the device generates those by sending up/down events with the correct timing. Relying on the jacdac bus packets is probably too brittle. Sending a click is very common.

services/mouse.md
pelikhan: yeah that's good

services/mouse.md
pelikhan: done

services/mouse.md
mmoskal: sounds good!

services/mouse.md
```diff
@@ -0,0 +1,42 @@
+# Mouse
+
+    identifier: 0x1885dc1c
+
+Control a device that acts as a mouse.
+
+## Commands
+
+    flags Button : u8 {
```
mmoskal, line 9: I think there might be mice with more than 8 buttons
```suggestion
    flags Button : u16 {
```

services/mouse.md
```diff
@@ -0,0 +1,42 @@
...
+        Down = 0x02
+        Click = 0x03
+        DoubleClick = 0x04
+    }
+    command set_button @ 0x80 {
+        buttons: Button
+        event: ButtonEvent
+    }
+
+Sets the up/down state of one or more buttons.
```
mmoskal, line 25: We should specify:

A `Click` is the same as `Down` followed by `Up` after 100ms.
A `DoubleClick` is two clicks with 150ms gap between them (that is, 100ms first click, 150ms gap, 100ms second click).

spectool/jdspec.ts
```diff
@@ -684,6 +684,10 @@ export function parseSpecificationMarkdownToJSON(filecontent: string, includes?:
             case "high":
                 info.highCommands = !!parseIntCheck(words[2])
                 break
+            case "status":
+                if (["stable", "experimental", "deprecated"].indexOf(words[2]) > -1)
+                    info.status = <any>words[2];
```
mmoskal, line 689: else error("unknown status")

pelikhan:
Commit this as is to test experimental rendering.

PR #46, opened 2020-10-28
keyboard spec

pelikhan:
HID keyboard spec

### DIFF
--- a/services/keyboard.md
+++ b/services/keyboard.md
@@ -1,0 +1,33 @@
+# Keyboard
+
+    identifier: 0x18b05b6a
+    
+A keyboard that can send keystrokes.
+
+The codes for the key (selectors) is defined in the [HID Keyboard
+specification](https://usb.org/sites/default/files/hut1_21.pdf), chapter 10 Keyboard/Keypad Page, page 81.
+    
+## Commands
+
+    flags Modifiers : u8 {
+        LeftControl = 0xe0
+        LeftShift = 0xe1
+        LeftAlt = 0xe2
+        LeftGUID = 0xe3
+        RightControl = 0xe0
+        RightShift = 0xe1
+        RightAlt = 0xe2
+        RightGUID = 0xe3
+    }
+    command key @ 0x80 {
+    repeats:
+        selector: u8
+        modifiers: u8
+        
+    }
+    
+Presses a key or a sequence of keys down.
+
+    command clear @ 0x82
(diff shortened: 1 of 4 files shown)


services/keyboard.md
```diff
@@ -0,0 +1,39 @@
...
+    identifier: 0x18b05b6a
+    
+A keyboard that can send keystrokes.
+
+The codes for the key (selectors) is defined in the [HID Keyboard
+specification](https://usb.org/sites/default/files/hut1_21.pdf), chapter 10 Keyboard/Keypad Page, page 81.
+    
+## Commands
+
+    flags Modifiers : u8 {
```
mmoskal, line 12: you would need to use different numbers (separate bits) to make this flags; given that there's already 8 of them, I suggest u16

services/keyboard.md
```diff
@@ -0,0 +1,39 @@
...
+        RightGUID = 0xe3
+    }
+    enum Action : u8 {
+        Press = 0
+        Up = 1
+        Down = 2
+    }
+    command key @ 0x80 {
+    repeats:
+        selector: u8
```
mmoskal, line 29: selector is 16 bit in the HID spec (though they only define codes until 0xE8)

services/keyboard.md
```diff
@@ -0,0 +1,39 @@
+# Keyboard
+
+    identifier: 0x18b05b6a
+    
+A keyboard that can send keystrokes.
+
+The codes for the key (selectors) is defined in the [HID Keyboard
+specification](https://usb.org/sites/default/files/hut1_21.pdf), chapter 10 Keyboard/Keypad Page, page 81.
```
mmoskal, line 8: I think we should repeat the list here...

services/keyboard.md
mmoskal: do they even treat modifiers separately in the HID spec? maybe we also don't have to

mmoskal: oh and the key is GUI not GUID

services/keyboard.md
pelikhan: it's like 5 pages long and anyone implementing HID will have to lookup that table.

services/keyboard.md
pelikhan: Ooops, copy paste issue. The HID spec reserves a single range for the modifiers so u8 is enough

PR #48, opened 2020-10-29
led matrix spec

pelikhan:


services/ledmatrix.md
```diff
@@ -0,0 +1,32 @@
...
+
+    rw enabled: boolean @ 0x81
+ 
+ Disables or enables the whole screen.
+ 
+    rw brightness: u8 @ 0x82
+    
+Sets the general brightness of the LEDs.
+ 
+    const rows: u8 # @ 0x83
```
pelikhan, line 22: u16

services/ledmatrix.md
```diff
@@ -0,0 +1,32 @@
...
+    
+Number of rows on the screen
+
+    const columns: u8 # @ 0x84
```
pelikhan, line 26: u16

services/ledmatrix.md
```diff
@@ -0,0 +1,32 @@
...
+    identifier: 0x1d35e393
+
+A rectangular matrix of monochrome LEDs
+
+## Registers
+
+    rw leds: bytes @ 0x80
+
+Read or writes the state of the screen where pixel information is 
+stored as bits, row by row.
```
pelikhan, line 12: Every row should be byte aligned.

pelikhan: Column first.

services/ledmatrix.md
```diff
@@ -0,0 +1,32 @@
+# LED Matrix
+
+    identifier: 0x1d35e393
```
pelikhan, line 3: experimental

services/ledmatrix.md
pelikhan:
```
ro leds: bytes @ reading
rw plot: bytes @ 0x82
```

pelikhan: rw leds: bytes @ 0x80 reading

pelikhan: import streaming_samples with the same logic as with sensor etc...

pelikhan: move streaming info to new service

PR #49, opened 2020-10-29
magnetometer spec

pelikhan:


services/magnetomer.md
```diff
@@ -0,0 +1,24 @@
...
+        x: i16.16 F
+        y: i16.16 F
+        z: i16.16 F
+    }
+
+Indicates the current magnetic field on magnetometer.
+
+## Commands
+
+    command calibrate @ calibrate
```
pelikhan, line 21: need a register to expose calibration status: calibration needed, calibrating, calibrated
+ event to notify that calibraiton changed

pelikhan: or send the "calibrated" register update when it changes

services/magnetomer.md
```diff
@@ -0,0 +1,24 @@
...
+    identifier: 0x13029088
+    extends: _sensor
+    camel: magneto
+
+A 3-axis magnetometer.
+
+## Registers
+
+    ro forces @ reading {
+        x: i16.16 F
```
mmoskal, line 12: shouldn't this be T (for Teslas)? I think we should go for nT and use an integer though (like codal) - eg the MAG3110 has range +-1,000,000nT and resolution of 100nT, so we're easily in the int range even for 100x better sensors and much stronger fields

pelikhan:
We need to chat about calibration today. it's a more complex state machine that just a command.

services/magnetomer.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Registers
+
+    ro forces @ reading {
+        x: i32 nT
+        y: i32 nT
+        z: i32 nT
+    }
+
+Indicates the current magnetic field on magnetometer.
```
mmoskal, line 17:
```suggestion
Indicates the current magnetic field on magnetometer.
For reference: `1 mgauss` is `100 nT` (and `1 gauss` is `100 000 nT`).
```

datasheets seem to talk in gauss not teslas (I'm sure they are against Elon!)

mmoskal (approved):

PR #53, opened 2020-11-11
Add C header files to dist

pelikhan:
Generate individual C header files in dist.


mmoskal:
I'm a bit of two minds on this - maybe we should just use the generated files as templates (that can and should be hand-edited); otherwise we will never cover all use cases; OTOH the constants file can be included in the user file, which is probably a good idea.

pelikhan:
The spec should allow to define additional constants?

mmoskal:
Let's see what's needed in jd-c

pelikhan:
I really think we should have one source of truth for the constants and as much as possible it should be the spec.

jamesadevine:
I think we should be going a step further and generating complete JACDAC interfaces, and maybe even the code for communicating with the underlying I2C peripherals (stretch goal).

mmoskal:
I find the fraction of code that can be generated for different services in jd-c to be very limited (other than constants and structs for binary layout).

The registers are currently defined semi-dynamically. The parsing code is 376 bytes and it handles read, write, sign and zero-extensions. I'm not sure generated code would be more efficient. One reason to do it differently would be for readability.

mmoskal (approved):

PR #57, opened 2020-11-12
Spec revisions

jamesadevine:


jamesadevine:
Where should "device identifiers" live?

jamesadevine:
Perhaps we need a devices page... We could include all the info about adding devices to the repository there too?

pelikhan:
We have a devices page in jacdac-ts... you can leave it as a separate file for now IMO and we'll update everything later on.

tballmsft:
We need to mention explicitly that the protocol does not do anything about allocation of device identifiers and have a little bit about it. The HDK needs to have a page about generating device identifiers, for sure.

PR #65, opened 2020-11-17
pressure button spec

pelikhan:
A button good enough to build a decent piano.


jamesadevine:
QQ: why do none of the button services emit the other button events such as click or long press? 

@mmoskal

mmoskal:
Laziness.

jamesadevine:
Haha, ok. I'll add them later on today.

pelikhan:
Time to test a base "button" sensor

pelikhan:
@mmoskal could this be subservice scenario, where we get a pressure reading as well?

mmoskal:
I can well imagine someone who really needs a pressure button service, and will not be happy with a binary button. OTOH in some scenarios you may be happy with either, but would like to use the pressure value if provided. I guess we should allow clients (and role mgr) binding to sub-services, then we can have the cake and eat it.

pelikhan:
@mmoskal  updated

services/analogbutton.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Registers
+
+    ro pressure: u0.16 / @ reading
+
+Indicates the current pressure (``force``) on the button.
+
+    rw low_threshold: u0.16 / @ low_threshold
+    
+Indicates the lower threshold and upper threshold for ``up`` event.
```
mmoskal, line 16: I'm wondering if we shouldn't use `active` / `inactive` for event names; now we have low threshold for up, and high for down, which seems... upside down. If it's high threshold for active it would make more sense.

mmoskal (approved):

PR #73, opened 2020-11-20
Migrate modules to JSON

pelikhan:


mmoskal:
I think you'll want to remove `source` from json files, and also add "Microsoft Research" or something as manufacturer.

PR #91, opened 2020-11-30
status code report

pelikhan:
Report for errors.


mmoskal (approved):

mmoskal:
Hmm... wait; it was supposed to be in _base not control?

pelikhan:
Right... moved.

pelikhan:
- rename _base to _system
- move status_code to _base
- implicit extends from _base for all services
- store class name derived - super.

pelikhan:
@mmoskal  moved away from enum in statuscode as we seem to duplicate it in every service.

PR #99, opened 2020-12-03
store pack info in packet

pelikhan:
@mmoskal would be good to store for the format info in the generated spec. I'm not sure about the format anymore. Could you finish "packFormat"?


pelikhan:
It's broken, but i'm using the format in the UI. Needs fixing.

PR #105, opened 2020-12-08
checking ranges

pelikhan:
is set_register and _get_register special and out of range?

The spec tool now checks for valid ranges of identifiers.

```
process _system.md
..\services\_system.md(26): get_register identifier 0x1000 out of range, expected in [0x0, 0x7f], [0x80, 0xeff], [0xf00, 0xfff]
..\services\_system.md(32): set_register identifier 0x2000 out of range, expected in [0x0, 0x7f], [0x80, 0xeff], [0xf00, 0xfff]
```


mmoskal:
yeah, get/set register are special and out of usual command range;

otherwise looks good!

PR #124, opened 2021-01-04
Add power negotiation protocol

mmoskal:
Could you guys read it and try to poke holes in it?


services/power.md
```diff
@@ -4,6 +4,35 @@
...
+The purpose of the power negotiation is to ensure that there is no more than ~500mA
+delivered to the power rail.
+This is realized by limiting the number of enabled power provider services to 1.
+
+Note, that it's also possible to have low-current power providers, which do not
+run a power provider service.
+These are **not** accounted for in the power negotiation protocol.
+
+* Upon reset, a power service enables itself.
+* Every enabled power service emits power `on` reports with its announce packets.
```
pelikhan, line 18: - wall vs battery: should we prioritize wall power from battery power?
- battery level: should we always pick the highest battery leve possible?

services/power.md
```diff
@@ -4,6 +4,35 @@
...
+
+The purpose of the power negotiation is to ensure that there is no more than ~500mA
+delivered to the power rail.
+This is realized by limiting the number of enabled power provider services to 1.
+
+Note, that it's also possible to have low-current power providers, which do not
+run a power provider service.
+These are **not** accounted for in the power negotiation protocol.
+
+* Upon reset, a power service enables itself.
```
pelikhan, line 17: does "enable" mean power the rail?

services/power.md
```diff
@@ -4,6 +4,35 @@
...
+* Every enabled power service emits power `on` reports with its announce packets.
+* If an enabled power service sees a power `on` report, it disables itself (unless in protection period).
+* If a power service has been disabled for at least 55-60s (random), it enables itself, sends
```
pelikhan, line 20: why 1 minute? just arbitrary?

services/power.md
pelikhan: does it enter in protection mode here too?

pelikhan:
What happens in the jacdac rail is directly powered from a power supply and we add a power module? It will not see the "on" report and will try to power the line?

services/power.md
mmoskal: yeah, thinking about this...

services/power.md
mmoskal: yes

mmoskal:
Directly powering JD rail without an MCU and a current limiter is not supported. It's also something we can't protect against (just like plugging it into 110V outlet).

services/power.md
mmoskal: the power service often wouldn't know the power source or capacity (it may be just a microUSB -> JACDAC "converter", which has no insight into where the power is coming from).

Still, I guess we could have some notion of priority.

pelikhan: If we envision power module with built-in lipo+charging circuit, one would expect those modules to receive power while developing (e.g. brain is connected to computer powering the bus) and give power when the brain is disconnected.

pelikhan:
> Directly powering JD rail without an MCU and a current limiter is not supported. It's also something we can't protect against (just like plugging it into 110V outlet).

We should explicitely mention this in the spec as a reminder.

pelikhan:
Low battery scenario:

Once a power supply is running out of battery, send a distress message to let other power supplies take over? This would allow to connect multiple batteries and depleate them one after the other.

pelikhan:
Rechaearging scenario:

Can a power supply use jacdac power to rechearge itself?

mmoskal:
> Once a power supply is running out of battery, send a distress message to let other power supplies take over? This would allow to connect multiple batteries and depleate them one after the other.

I think this already works - when you're low on battery, you don't turn yourself on the 1 minute interval - this will have others take up the load. If you're not low on battery, the load will be shared, minute by minute.

Again, if the power source if a microUSB you don't know how much battery is left. In that case, the battery is going to run out, and whole system will shut down and resume after ~1s. If you have two power providers, with similar battery capacities, that should only happen when the other one is also almost empty.

mmoskal:
> Can a power supply use jacdac power to rechearge itself?

Talked with @pelikhan offline - decided to have another service, which turns on charging. Automatically charging is likely to overload the bus (eg if more than one battery wants to charge, or you already have lights connected to the bus drawing close to 500mA).

services/power.md
mmoskal: OK, not doing this in this service

services/power.md
mmoskal: added info

services/power.md
```diff
@@ -4,6 +4,69 @@
 
 A power-provider service.
 
+## Power negotiation protocol
+
+The purpose of the power negotiation is to ensure that there is no more than ~500mA
+delivered to the power rail.
+This is realized by limiting the number of enabled power provider services to 1.
```
pelikhan, line 11: 1 --> **one**

services/power.md
```diff
@@ -4,6 +4,69 @@
...
+
+Note, that it's also possible to have low-current power providers, which do not
```
pelikhan, line 13: link to more info about "low-current power providers" ?

services/power.md
```diff
@@ -4,6 +4,69 @@
...
+The protocol is based on `on` reports, which are always sent 
+after general device announce packets, in the same frame.
+This makes it simpler for other power services to parse them.
+
+The `on` reports contain device priority, which is either its remaining battery
+capacity, or a constant.
+
+After queuing an announce with `on` report, the service enters a grace period
+until the report has been sent on the wire.
+During grace period incoming `on` reports are ignored.
```
pelikhan, line 26: During **the**

services/power.md
```diff
@@ -4,6 +4,69 @@
...
+
+* Upon reset, a power service enables itself (instant-on), and after 0-300ms (random)
+  send the first announce packet (with `on` report)
```
pelikhan, line 29: so it sends it twice?

services/power.md
```diff
@@ -4,6 +4,69 @@
...
+* Every enabled power service emits power `on` reports with its announce packets,
```
pelikhan, line 30: maybe the "on" report should be renamed to "enabled" or "active" to be less confusing with 'on' word.

services/power.md
```diff
@@ -4,6 +4,69 @@
...
+  which are sent every 400-600ms (random; first few announce packets can be even sent more often)
+* If an enabled power service sees a power `on` report from somebody else of higher or equal priority,
```
pelikhan, line 32: priority has not been introducted yet here.

jamesadevine:
What does "send the first announce packet (with `on` report)" mean?

Is this something that is contained in the control layer announce packet, or does this refer to the power service's own announce packet?

services/power.md
mmoskal: no link yet; added info

services/power.md
mmoskal: no; cleared up

services/power.md
mmoskal: it's defined a few lines above

PR #137, opened 2021-01-13
servo

pelikhan:
- [x] servo service should use angles and hide the PWM details in the firmware implementation.
- [x] servo tend to drift; hence it's always good to have a correction parameter to "calibrate" them.
Requires firmware update.
- [x] report stall torque (basically size of servo) and rotation speed


services/servo.md
```diff
@@ -2,15 +2,18 @@
...
     identifier: 0x12fc9103
 
-Servo is a small motor directed with a PWM signal.
-This services fixes the servo period at 20ms, and the pulse can be regulated.
+Servo is a small motor with arm that can be pointing at a specific direction.
 
 ## Registers
 
-    rw pulse: u32 us {typical_min = 500, typical_max = 2500} @ value
+    rw angle: i16 ° {typical_min = -90, typical_max = 90} @ value
```
mmoskal, line 9: I would be inclined to use `deg` as the unit name; I think we use ASCII names everywhere

services/servo.md
```diff
@@ -2,15 +2,18 @@
...
-Specifies length of the pulse in microseconds. The period is always 20ms.
+Specifies the angle of the arm.
 
     rw enabled: bool @ intensity
 
-Turn the power to the servo on/off.
\ No newline at end of file
+Turn the power to the servo on/off.
+
+    rw drift?: i16 ° @ 0x81
```
mmoskal, line 17: how about `drift_left` `drift_mid` `drift_right` ?

services/servo.md
mmoskal: I think we can have more resolution than 1 deg (the previous one had 1us out of ~1000, so ~0.2deg). Maybe we should use `i16.16` type ? (both here and for drift)

pelikhan: I'll address the ° issue in a later PR. It's currently used in the display which should be °.

services/servo.md
pelikhan: Not sure how you apply those... so i'll let you add them.

services/servo.md
```diff
@@ -2,15 +2,47 @@
...
+
+Lowest angle that can be set.
+
+    const max_angle?: i16.16 ° @ max_reading
+
+Highest angle that can be set.
+
+    enum Variant: u32 {
+        PositionalRotation = 1,
+        ContinuousRotation = 2,
```
mmoskal, line 31: should we drop the continuous servo here? it should use the motor service (in a given project, you can't really replace positional servo with a continuous one, but you can replace a DC motor with a continuous servo)

services/servo.md
```diff
@@ -2,15 +2,47 @@
...
+        Linear = 3,
+    }
+    const variant: Variant @ variant
+
+Specifies the type of thermometer.
```
mmoskal, line 36:
```suggestion
Specifies the type of servo.
```

services/servo.md
```diff
@@ -2,15 +2,45 @@
...
     identifier: 0x12fc9103
 
-Servo is a small motor directed with a PWM signal.
-This services fixes the servo period at 20ms, and the pulse can be regulated.
+Servo is a small motor with arm that can be pointing at a specific direction.
 
 ## Registers
 
-    rw pulse: u32 us {typical_min = 500, typical_max = 2500} @ value
+    rw angle: i16.16 ° { typical_min=-90, typical_max=-90 } @ value
```
mmoskal, line 9:
```suggestion
    rw angle: i16.16 ° { typical_min=-90, typical_max=90 } @ value
```

PR #139, opened 2021-01-13
Add min/max_reading, reading_error, and variant

### LINKED ISSUE #134: variant optional register
variant register allows for a module to clarify the hardware variant of a generic service.

For example a potentiometer would have a "slider" and "rotary" variants.

```
     const variant @ base
```

In service, it can be specilized with an enum.

### LINKED ISSUE #135: Thermometer does not specify temperature range
There should be register to specify the value range. Thermometer can have vastly different operating temperatures. This is true for any sensor really.

mmoskal:
Fixes #134
fixes #135 

(slider should be renamed, but leaving this out for now)


services/_system.md
```diff
@@ -85,6 +85,19 @@ Period between packets of data when streaming in milliseconds.
...
 
+    const min_reading: i32 @ 0x104
+
+The lowest value that can be reported by the sensor.
+
+    const max_reading: i32 @ 0x105
+
+The highest value that can be reported by the sensor.
+
+    ro reading_error: i32 @ 0x106
```
pelikhan, line 96: sensors typically use accuracy, eg. % of full scale value. should we use this instead?

pelikhan: This page is an interresting read: https://www.ni.com/en-us/innovations/white-papers/13/sensor-terminology.html#:~:text=Back%20to%20top-,Accuracy,scale%20or%20in%20absolute%20terms.

mmoskal: At least for temp sensors, it's absolute (0.2C), though it's bigger at the end of the range. Similar for humidity. I guess % is equivalent but absolute values are more clear-defined.
Screen Shot 2021-01-13 at 2 23 22 PM

pelikhan: maybe we can change this register to return a sequence of points and we do a linear interpoluation of the error.

mmoskal: It should give you the current error. So in the example above, if it reads -20C it should return 0.5C error, and if it reads 25C it should return 0.2C error. The curves can have any shape I think.

services/_system.md
```diff
@@ -104,6 +117,11 @@ add this report in frame along with the announce packet.
 
 Preferred default streaming interval for sensor in milliseconds.
 
+    const variant: i32 @ 0x107
```
pelikhan, line 120: u32?

pelikhan (approved):

services/_system.md
mmoskal:
```suggestion
    const variant: u32 @ 0x107
```

pelikhan (commented):
I'll rename the slider and everything related to that in another PR.

pelikhan:
LGTM

PR #161, opened 2021-01-18
add barometer spec

pelikhan:


services/barometer.md
```diff
@@ -0,0 +1,18 @@
...
+    identifier: 0x1e117cea
+    extends: _sensor
+
+A sensor measuring air pressure of outside environment.
+
+## Registers
+
+Default streaming interval is 1s.
+
+    ro pressure: u22.10 hPa { absolute_min = 300, absolute_max = 1100, typical_min = 940, typical_max = 1040 } @ reading
```
mmoskal, line 12: should absolute min/max be much wider? eg if someone uses this on Mars? maybe we should drop the absoulte

pelikhan: I was looking at some regular athmospheric sensors and those were the extremes supported. There is a whole other range of "pressure sensors" that support way more extreme ranges; i think we should keep them separate.

mmoskal: cool, let's keep the Mars use case for later!

PR #163, opened 2021-01-20
distance sensor

pelikhan:
Includes sonar/ir sensors.

### DIFF
--- a/services/distance.md
+++ b/services/distance.md
@@ -1,0 +1,30 @@
+# Distance
+
+A sensor that determines the distance of an object without any physical contact involved.
+
+        identifier:  0x141a6b8a
+        extends: _sensor
+
+## Registers
+
+    ro distance: u6.10 m { typical_min=0.02, typical_max=4 } @ reading
+
+Current distance from the object
+
+    const min_range?: u6.10 m @ min_reading
+
+Minimum measurable distance 
+
+    const max_range?: u6.10 m @ max_reading
+
+Maximum measurable distance
+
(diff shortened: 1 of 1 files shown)


services/distance.md
```diff
@@ -0,0 +1,30 @@
+# Distance
+
+A sensor that determines the distance of an object without any physical contact involved.
+
+        identifier:  0x141a6b8a
+        extends: _sensor
+
+## Registers
+
+    ro distance: u6.10 m { typical_min=0.02, typical_max=4 } @ reading
```
mmoskal, line 10: maybe we should standardize on u16.16 when there are no good to reasons to do otherwise? there is almost no transfer cost difference (and often literally none as packets in frames are padded to full word) and it gives us more range and precision.

services/distance.md
```diff
@@ -0,0 +1,30 @@
...
+
+    const max_range?: u6.10 m @ max_reading
+
+Maximum measurable distance
+
+    enum Variant: u32 {
+        Ultrasonic = 1,
+        Infrared = 2,
+        LiDAR = 3,
+        TimeOfFlight = 4,
```
mmoskal, line 26: is time-of-flight laser?

pelikhan: Yes it's laser.

services/distance.md
mmoskal: fixed

services/distance.md
mmoskal: renamed

PR #167, opened 2021-01-21
relay service

pelikhan:


services/relay.md
```diff
@@ -0,0 +1,30 @@
+# Relay
+
+    identifier: 0x183fe656
+
+A switching relay.
+
+## Registers
+
+    rw active: bool @ intensity
```
pelikhan, line 9: should this be value?

mmoskal: No. Maybe cal it "conducting" ?

pelikhan: "close" ?

mmoskal: I think "closed" may be interpreted the other way by people not familiar with electric circuits; conducting should be clear (if they know the word)

pelikhan: conductive it is

mmoskal:
you could also add registers for maximum current and voltage

mmoskal (approved):

services/relay.md
```diff
@@ -0,0 +1,34 @@
...
+
+    event on @ active
+
+Emitted when relay goes from ``off`` to ``on`` state.
+
+    event off @ inactive
+
+Emitted when relay goes from ``on`` to ``off`` state.
+
+    const max_switching_current?: u16 A @ 0x180
```
mmoskal, line 32: this register is under events; I guess we should have the spectool check for that
also I think u16.16 would be better here

PR #168, opened 2021-01-21
character lcd screen

pelikhan:


services/characterlcdscreen.md
```diff
@@ -0,0 +1,49 @@
+# Character LCD Screen
```
mmoskal, line 1: alphanumeric display? or character display? 

I would drop LCD, since it may as well be LED display. I could also imagine implementing this on a 5x5 microbit style display

services/characterlcdscreen.md
```diff
@@ -0,0 +1,49 @@
...
+
+    identifier: 0x1f37c56a
+
+A character LCD screen.
+
+## Registers
+
+    rw message: string @ value
+
+Sets text, with `\n` as the newline character.
```
mmoskal, line 11:
```suggestion
Text to show. Use `\n` to break lines.
```

services/characterlcdscreen.md
```diff
@@ -0,0 +1,49 @@
...
+
+    rw blink: bool @ 0x81
+
+Configures to blink the cursor.
+
+    enum TextDirection : u8 {
+        LeftToRight = 1,
+        RightToLeft = 2
+    }
+    rw text_direction: TextDirection @ 0x82
```
mmoskal, line 25: Should be optional, for screens that don't have Hebrew/Arabic font anyways.

services/characterlcdscreen.md
```diff
@@ -0,0 +1,49 @@
...
+
+Specifies the RTL or LTR direction of the text.
+
+    rw backlight_color? @ 0x83 {
+        red: u8
+        green: u8
+        blue: u8
+    }
+
+If available, sets the backlight color.
```
mmoskal, line 35: should we do HSV, so that it's clear V is used when backlight is monochromatic?

services/characterlcdscreen.md
```diff
@@ -0,0 +1,49 @@
...
+
+    const rows?: u8 @ 0x180
+
+Gets the number of rows. Default is ``2``.
```
mmoskal, line 39: I don't think this should be optional - I don't see why one would skip it. 

Also, you can use syntax `rows = 2: u8` to specify value at reset, but I guess we should now do that for const registers (since they never change)

services/characterlcdscreen.md
```diff
@@ -0,0 +1,49 @@
...
+
+    const columns?: u8 @ 0x181
+
+Gets the number of columns. Default is ``16``.
```
mmoskal, line 43: same as for rows.

services/characterlcdscreen.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Commands
+
+    command clear @ 0x80 { }
+
+Clears the screen
```
mmoskal, line 49: is this different than setting text to `""` ? if the same, I think we should skip it

mmoskal (commented):
wow, that's lots of comments ;))

services/characterlcdscreen.md
pelikhan: I was thinking about those specifically: https://learn.adafruit.com/character-lcds/overview

pelikhan: I think the 5x5 LED screen is something else

services/characterlcdscreen.md
mmoskal: I guess this is backlight color or LED color, in case it's an LED display

pelikhan: Something like this would allow me to mount better UIs with color pickers
```
rw backlight_color: u24 hsv @ 0x180
```

mmoskal: I think we could use `hsv` as type; we don't really have `u24` type.

services/characterlcdscreen.md
mmoskal: ok, ignoring the 5x5 scrolling display, there are character displays that are LEDs (eg on buses; i'm sure there are other examples); also even if it's a graphical display, it could still expose the character service

pelikhan:
Character screen? I think LCD really helps discover it. There are tons of models but I haven’t spotted a pure LED one.

mmoskal:
I think an OLED screen module should implement this service (in addition to some sort of graphics service).
It's different to the LCD character screen, but in some sense OLED pixel screen is strictly better, so we can implement it.
We could also have the OLED screen implement 5x5 micro:bit screen.

This way your project will work with whatever screen you have.

services/characterlcdscreen.md
pelikhan: ok, i'll add back backlight once we have a hSV type.

PR #175, opened 2021-01-21
improve vibration

pelikhan:
Vibration firmware should have a concept of "auto-off"


services/vibration.md
```diff
@@ -10,3 +10,12 @@ Control a vibration motor.
 
 Rotation speed of the motor.
 If only one rotation speed is supported, then `0` shell be off, and any other number on.
+
+## Commands
+
+    command vibrate @ 0x80 {
```
mmoskal, line 16: maybe we should really do vibration patterns, like you said before...

pelikhan: yes!

pelikhan: and it should be finite. you don't want to have it vibrate forever becaues of a bug

pelikhan:
```
{
repeats:
    intensity: u8 /
    duration: u8 8ms
}
```

pelikhan:
@mmoskal I tweaked the service. speed is a reading, enabled is on/off and you have to use commands to launch "vibration" nimations.

mmoskal (approved):

PR #177, opened 2021-01-21
Service: Wind direction

pelikhan:

### DIFF
--- a/services/winddirection.md
+++ b/services/winddirection.md
@@ -11,10 +11,10 @@ A sensor that measures wind direction.
 
 The direction of the wind.
 
-    rw direction_offset?: i16 ° @ 0x80
+    ro wind_direction_offset?: i16 ° @ 0x80
 
 Offset added to direction to account for sensor calibration.
 
-    const direction_resolution?: u16 ° @ reading_resolution
+    ro wind_direction_error?: u16 ° @ reading_error
 
-Small change in direction that can be detected by the sensor.
+Error on the wind direction reading
--- a/services/winddirection.md
+++ b/services/winddirection.md
@@ -14,3 +14,7 @@ The direction of the wind.
     rw direction_offset?: i16 ° @ 0x80
 
 Offset added to direction to account for sensor calibration.
+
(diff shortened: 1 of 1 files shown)


services/winddirection.md
```diff
@@ -0,0 +1,20 @@
+# Wind direction
+
+A sensor that measures wind direction.
+
+    identifier: 0x186be92b
+    extends: _sensor
+
+## Registers
+
+    ro wind_direction: u16 ° @ reading
```
mmoskal, line 10: maybe overkill but u16.16 for future-proofing?

services/winddirection.md
```diff
@@ -0,0 +1,20 @@
...
+
+The direction of the wind.
+
+    rw direction_offset?: i16 ° @ 0x80
+
+Offset added to direction to account for sensor calibration.
+
+    const direction_resolution?: u16 ° @ reading_resolution
```
mmoskal, line 18: should it also have error?

services/winddirection.md
pelikhan: does not hurt... but the spark fun sensor precision was around 40deg

services/winddirection.md
pelikhan: yeah they come in pair

pelikhan: use error,. drop resolution

services/winddirection.md
```diff
@@ -0,0 +1,20 @@
...
+    identifier: 0x186be92b
+    extends: _sensor
+
+## Registers
+
+    ro wind_direction: u16 ° @ reading
+
+The direction of the wind.
+
+    ro wind_direction_offset?: i16 ° @ 0x80
```
mmoskal, line 14: what is this one? is it somehow detected by the sensor?

pelikhan:
The idea of the offset would be to calibrate the direction based on the orientation of the sensor

services/winddirection.md
mmoskal: > The idea of the offset would be to calibrate the direction based on the orientation of the sensor

but it's read-only, I'm not sure how would that work

PR #178, opened 2021-01-21
Service: Light level

pelikhan:

### DIFF
--- a/services/lightlevel.md
+++ b/services/lightlevel.md
@@ -1,0 +1,21 @@
+# Light level
+
+An analog sensor that measures luminosity level.
+
+    identifier: 0x17dc9a1c
+    extends: _sensor
+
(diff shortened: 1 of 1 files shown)


services/lightlevel.md
```diff
@@ -0,0 +1,21 @@
+# Light level
```
mmoskal, line 1: I suppose we would have a separate luxmeter service for a precise one?

pelikhan: yes luxmeter would be a different one, much more high-end

services/lightlevel.md
```diff
@@ -0,0 +1,21 @@
...
+
+An analog sensor that measures luminosity level.
```
pelikhan, line 3: analog -> shitty

PR #179, opened 2021-01-21
Service: Equivalent CO²

pelikhan:

### DIFF
--- a/services/eco2.md
+++ b/services/eco2.md
@@ -8,18 +8,22 @@ Measures equivalent CO² levels.
 
 ## Registers
 
-     ro e_CO2: u16 ppm { typical_min=400, typical_max=8192 }  @ reading
+     ro e_CO2: u22.10 ppm { typical_min=400, typical_max=8192 }  @ reading
 
 Equivalent CO² (eCO²) readings.
 
-    const min_e_CO2?: u16 ppm @ min_reading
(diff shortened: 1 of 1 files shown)


services/eco2.md
```diff
@@ -0,0 +1,25 @@
...
+
+Measures equivalent CO² levels.
+
+    identifier: 0x169c9dc6
+    extends: _sensor
+    camel: eCO2
+
+## Registers
+
+     ro e_CO2: u16 ppm { typical_min=400, typical_max=8192 }  @ reading
```
mmoskal, line 11: should we do u32? or maybe u22.10? in general, there isn't much advantage using u16/u8 readings for sensors given all the other overheads in the system

pelikhan: use u32

pelikhan: use u22.10

services/eco2.md
```diff
@@ -0,0 +1,25 @@
...
+
+    const min_e_CO2?: u16 ppm @ min_reading
+
+Minimum measurable value
+
+    const max_e_CO2?: u16 ppm @ max_reading
+
+Minimum measurable value
+
+    const conditioning_period?: u16 min @ 0x180
```
pelikhan, line 23: use seconds

pelikhan: u32 it's a const anyway

PR #187, opened 2021-01-22
Service: Gyroscope

pelikhan:


services/gyroscope.md
```diff
@@ -0,0 +1,29 @@
...
+        z: i12.4 °/s
+    }
+
+Indicates the current forces acting on accelerometer.
+
+    ro rotation_rates_error: i12.4 @ reading_error
+
+Error on the reading value.
+
+    enum RotationRateRange: u8 {
```
pelikhan, line 22: we should add a range to the accelerometer too

services/gyroscope.md
```diff
@@ -0,0 +1,29 @@
...
+        Slow = 1,
+        Medium = 2,
+        Fast = 2
+    }
+    rw rotation_rate_range?: RotationRateRange @ 0x80
+
+Configures the range of range of rotation rates. Typically, ``slow`` is ``250 °/s``, ``medium``, is ``500 °/s`` and ``fast`` is ``2000 °/s``.
```
mmoskal, line 29: let's make it into `rw max_rate: i16.16 deg/s`

The device should set the lowest rate that is not smaller than the one requested (eg., if 250 and 500 are available then setting to 0-250 sets it to 250, and setting to 251 or more sets it to 500).
Reading back the rate can give you the actual value.

mmoskal (approved):

PR #190, opened 2021-01-22
Service: Line tracking

pelikhan:


services/linetracking.md
```diff
@@ -0,0 +1,31 @@
...
+A sensor that detects light and dark surfaces, typically used for line following robots.
+
+    identifier: 0x126c4cb2
+    extends: _sensor
+
+## Registers
+
+    ro brightness: u8 / @ reading
+
+Reports the detected brightness. It may be a digital value or, for some sensor, analog value.
```
mmoskal, line 12: Let's go for u16 - most ADCs are more than 8 bit.

Also what do you mean digital or analog value?

services/linetracking.md
```diff
@@ -0,0 +1,31 @@
...
+
+    enum Variant: u8 {
+        InfraredDigital = 1,
+        InfraredAnalog = 2,
+        InfraredAnalog5ch = 3
+    }
+    const variant?: Variant @ variant
```
mmoskal, line 19: again, not sure what is analog/digital here

services/linetracking.md
```diff
@@ -0,0 +1,31 @@
...
+
+Type of physical sensor used
+
+## Events
+
+    event dark @ 0x80 {}
+
+The sensor detected a transition from light to dark
```
mmoskal, line 27: you could use @ active/inactive (doesn't matter really which way) - this way there is already built-in tracking of status in the client

services/linetracking.md
pelikhan: yes. most sensor reports high/low but some have more sensors and are better suited for PID control

services/linetracking.md
pelikhan: analog output

mmoskal: oh, so digital one would report either 0 or 255. got it.

mmoskal: the 5ch one - I guess it would tell you where the line is (ie the value at each of the 5 channels), wouldn't it?

PR #191, opened 2021-01-24
Service: Real time clock

pelikhan:

Leaving the alarm stuff aside for now.


services/realtimeclock.md
```diff
@@ -0,0 +1,33 @@
...
+        year: u16
+        month: u8
+        day: u8
+        day_of_week: u8
+        hour: u8
+        min: u8
+        sec: u8
+    }
+
+Current time in 24h representation. Default streaming period is 1 second.
```
mmoskal, line 19: specify ranges - is month 1-12 or 0-11? similar for day, day of week; also say hour is 0-24 (I assume), min and sec 0-59 (I don't think we'll do leap seconds would we?)

we also need to say if, once set, the leap years are automatically detected (I guess yes)

services/realtimeclock.md
```diff
@@ -0,0 +1,33 @@
...
+         year: u16
+         month: u8
+         day: u8
+         day_of_week: u8
+         hour: u8
+         min: u8
+         sec: u8
+    }
+
+Sets the current day time on the clock.
```
mmoskal, line 33: What happens if you specify wrong day of week? I guess it's corrected.

Should this also have an optional command to "auto-set" time, with a given time zone? eg we could have this service on a GPS

mmoskal (approved):

services/realtimeclock.md
```diff
@@ -0,0 +1,33 @@
+# Real time clock
+
+    
+    identifier: 0x1a8b1a28
+    extends: _sensor
```
mmoskal, line 5: we should also have precision register, in parts per million, u16.16 (this is not error though)

services/realtimeclock.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Registers
```
mmoskal, line 7: we could also have error register, which counts the time since last `set_time` and based on precisions computes possible error in seconds

services/realtimeclock.md
```diff
@@ -0,0 +1,33 @@
...
+
+    ro now @ reading {
```
mmoskal, line 9: do we also want a reading as seconds since epoch? I guess we then need to synchronize between the two readings. not sure about that one

services/realtimeclock.md
pelikhan: I've added a few ranges on the register.

services/realtimeclock.md
pelikhan: Probably something we could add later.

services/realtimeclock.md
pelikhan: wrong service for this comment?

services/realtimeclock.md
pelikhan: would that be for low quality clocks?

services/realtimeclock.md
pelikhan: Most of these clocks use Y2K as the base time; we could add an offset to meet epoch - but since it's not going to be milliseconds, i taught it was fine. Maybe ``now`` is not a good name.

services/realtimeclock.md
pelikhan: Come to think of it, setting the time could be moved into a mixin service because it is hardware specific.

services/realtimeclock.md
mmoskal: even the high-precision one from adafruit https://www.adafruit.com/product/3013 is 2ppm which is 5s/month; the not so good ones are 2s/day

services/realtimeclock.md
mmoskal: no, for this one; DS3231 is 2ppm; crystals are often ~10ppm depending on temp

services/realtimeclock.md
mmoskal: yeah, let's skip the auto-set of time

services/realtimeclock.md
mmoskal: let's do 0-60 on seconds to allow for leap seconds

services/realtimeclock.md
mmoskal: I was thinking of another integer register; I'm fine with y2k. it would be mostly useful to measure passage of time - eg your JD device goes to sleep, wakes up, and you want to know how long you were out - computing date->seconds is quite complicated, so it would be nice if the service gave a monotonic second counter

services/realtimeclock.md
```diff
@@ -1,31 +1,28 @@
...
 Current time in 24h representation. Default streaming period is 1 second.
-Day of week starts on Mondays.
 
 ## Commands
 
     command set_time @ 0x80 {
          year: u16
          month: u8
          day: u8
-         day_of_week: u8
```
mmoskal: let's replace it with padding, so the structure is the same

services/realtimeclock.md
pelikhan: we have error, resolution as system registers, precision added?

mmoskal: in some sense the "precision" here is "rate of error"; it typically doesn't make sense to state it, but here it accumulates with time; I wouldn't add anything to system just yet

services/realtimeclock.md
pelikhan: we could ditch this blow-out reading and use epoch always, let the clients do the math.

mmoskal: we could, but the math is somewhat involved, with leap years and the like; and the hardware typically uses y/m/d/h/m/s, just like you do here.

services/realtimeclock.md
mmoskal: btw, what do we do about DST? nothing?

services/realtimeclock.md
mmoskal: I just learned there are atomic clocks (microchip sells one for $5k; it's only a few cm in size) that have precision of 0.00005 ppm. So I guess precision as u16.16 ppm would work - this from 0.00002 ppm to 6% (I don't think you could call something with 6% precision a "real time" clock).

PR #193, opened 2021-01-25
Service: Color

pelikhan:

### DIFF
--- a/services/color.md
+++ b/services/color.md
@@ -1,0 +1,16 @@
+# Color
+
+Senses RGB colors
+
+    identifier: 0x1630d567
+    extends: _sensor
+
+## Registers
+
+    ro color @ reading {
+        red: u8
+        green: u8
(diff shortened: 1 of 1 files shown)


services/color.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Registers
+
+    ro color @ reading {
+        red: u8
+        green: u8
+        blue: u8
+    }
+
+Detected color in the RGB color space.
```
mmoskal, line 16: maybe do `u16 /` ?

pelikhan: 16 bits for each color channels?

mmoskal: yes; there are these color sensors for calibrating photo machines, I suspect they are more than 8 bit per channel

pelikhan: fixed.

mmoskal (approved):

PR #194, opened 2021-01-25
Service: Illuminance

pelikhan:

### DIFF
--- a/services/illuminance.md
+++ b/services/illuminance.md
@@ -7,10 +7,10 @@ Measures light in Lux
 
 ## Registers
 
-    ro light: u16.16 lux { typical_max=100000 } @ reading
+    ro light: u22.10 lux { typical_max=100000 } @ reading
 
 The amount of illuminance, as lumens per square metre.
 
-    ro light_error?: u16.16 lux @ reading_error
+    ro light_error?: u22.10 lux @ reading_error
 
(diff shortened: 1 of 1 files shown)


services/illuminance.md
```diff
@@ -0,0 +1,12 @@
+# Illuminance
+
+Measures light in Lux
+
+    identifier: 0x1e6ecaf2
+    extends: _sensor
+
+## Registers
+
+    ro light: u16.16 lux { typical_min=0, typical_max=6555 } @ reading
```
mmoskal, line 10: I would drop the typical min (it's absolute from type)

Bright day outside is 100k lux; maybe that should be typical max? (my handheld light sensor can give readings in that range)

services/illuminance.md
```diff
@@ -0,0 +1,16 @@
+# Illuminance
+
+Measures light in Lux
+
+    identifier: 0x1e6ecaf2
+    extends: _sensor
+
+## Registers
+
+    ro light: u16.16 lux { typical_max=100000 } @ reading
```
mmoskal, line 10: whoops; 100k doesn't fit in u16 (added a TODO to check this in spectool)

should we go u22.10 ?

pelikhan: updated

mmoskal (approved):

PR #195, opened 2021-01-25
Service: Joystick

pelikhan:

### DIFF
--- a/services/joystick.md
+++ b/services/joystick.md
@@ -13,6 +13,7 @@ A two axis directional joystick
     }
 
 The direction of the joystick measure in two direction.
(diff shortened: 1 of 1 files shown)


services/joystick.md
```diff
@@ -0,0 +1,24 @@
...
+    extends: _sensor
+
+## Registers
+
+    ro direction @ reading {
+        x: i16 /
+        y: i16 /
+    }
+
+The direction of the joystick measure in two direction.
```
mmoskal, line 15:
```suggestion
The direction of the joystick measure in two direction.
If joystick is digital, then each direction will read as either `-0x8000`, `0x0`, or `0x7fff`.
```

mmoskal (approved):

PR #197, opened 2021-01-25
Service: Pulse Oximeter

pelikhan:


services/pulseoximeter.md
```diff
@@ -0,0 +1,18 @@
...
+**JACDAC is not suitable for medical devices and should NOT be used in any kind of device to diagnose or treat any medical conditions.**
+
+    identifer: 0x14ad1a5d
+    extends: _sensor
+
+## Registers
+
+    ro oxygen: i8.8 bpm @ reading
+
+The estimated oxygen level in blood. Negative value when the sensor cannot sense the rate.
```
mmoskal, line 14: This value is given as %, so maybe just `u16 /` ?

As in heart rate, 0 could be used to indicate an error (the sensors typically go 75-99%).

pelikhan: Actually we have a status code register on every service...

pelikhan: oops bpm should be %

mmoskal: wait, so you want it to be 100.0 for full saturation or 0xffff ? I guess 100.0 would follow relative humidity

pelikhan: i have a % type now.

mmoskal (approved):

PR #199, opened 2021-01-25
Service: Speech synthesis

pelikhan:
This pull request adds a new service definition for JACDAC. Allows a jacdac device to generate speech in the browser for example.

### DIFF
--- a/services/speechsynthesis.md
+++ b/services/speechsynthesis.md
@@ -1,0 +1,43 @@
+#  Speech synthesis
+
+     identifier: 0x1204d995
+
+A speech synthesizer
+
+## Registers
+
+    ro speaking: bool @ reading
+
+Reports if the engine is speaking.
+
+    rw enabled: bool @ intensity
+
+Determines if the speech engine is in a non-paused state.
+
+    rw lang?: string @ 0x80
+
+Default language used for utterances.
+
+    rw volume?: u8 / @ 0x81
+
+Default volume for utterances
+
+    rw pitch?: u8 / @ 0x82
+
+Default pitch for utterances
+
+    rw rate?: u8 / @ 0x83
+
+Default rate for utterances
+
+## Commands
+
+    command speak @ 0x80 {
+        text: string
+    }
+
+Adds an utterance to the utterance queue; it will be spoken when any other utterances queued before it have been spoken.
+
+    command cancel @ 0x81 {}
+
+Cancels all utterances from the utterance queue.


jamesadevine (approved):
Looks good, I think we're going to have to consider alternative languages as time goes on... e.g. string might not be suitable for mandarin example. 

Is the idea we would always send commands in english and the browser would perform conversion to alternate languages?

pelikhan:
The user can specify the locale using ``lang``, then use unicode characters in the utf8-encoded string to speak it out.

jamesadevine:
Sounds great, thanks Peli

services/speechsynthesis.md
```diff
@@ -0,0 +1,43 @@
...
+
+Reports if the engine is speaking.
+
+    rw enabled: bool @ intensity
+
+Determines if the speech engine is in a non-paused state.
+
+    rw lang?: string @ 0x80
+
+Default language used for utterances.
```
mmoskal, line 19: We should say how to specify it, I would say https://www.ietf.org/rfc/bcp/bcp47.txt (which is what HTML and XML use)

We could also rename this to `locale` (eg. en-us vs en-gb)

I would drop "Default", as I see no other way to change it

services/speechsynthesis.md
```diff
@@ -0,0 +1,43 @@
...
+
+    rw volume?: u8 / @ 0x81
+
+Default volume for utterances
```
mmoskal, line 23: I assume this typically doesn't start at 0? we could use `volume? = 128` syntax

services/speechsynthesis.md
```diff
@@ -0,0 +1,43 @@
...
+
+    rw pitch?: u8 / @ 0x82
```
mmoskal, line 25: should specify default for pitch

services/speechsynthesis.md
```diff
@@ -0,0 +1,43 @@
...
+
+Default pitch for utterances
+
+    rw rate?: u8 / @ 0x83
```
mmoskal, line 29: should specify default for rate

services/speechsynthesis.md
```diff
@@ -0,0 +1,43 @@
...
+
+    command speak @ 0x80 {
+        text: string
+    }
+
+Adds an utterance to the utterance queue; it will be spoken when any other utterances queued before it have been spoken.
+
+    command cancel @ 0x81 {}
+
+Cancels all utterances from the utterance queue.
```
mmoskal, line 43: does it also stop the one currently being spoken? (I would say yes)

services/speechsynthesis.md
pelikhan: somehow ``rw pitch? = 0.5:`` ... does not parse

PR #202, opened 2021-01-26
Service: DMX

pelikhan:

A one-way DMX controller; would be useful to drive many different lighting/show system.


services/dmx.md
```diff
@@ -0,0 +1,30 @@
...
+If available, terminate the DMX circuit.
+
+## Commands
+
+    command send @ 0x80 {
+        packet: pipe
+    }
+    pipe report packet {
+        channels: bytes
+    }
```
mmoskal, line 28: I think you need this:

```
    command send # 0x80 {}
    report {
        data_port: pipe_pore
    }
    pipe command data_port {
        data: bytes
    }
```

not sure if it makes sense to specify "format" of `data_port`, since it's just bytes.

services/dmx.md
```diff
@@ -0,0 +1,30 @@
...
+
+Send a DMX packet, up to 512bytes long.
```
mmoskal, line 30: I think it should be up to 513, including the start code

services/dmx.md
```diff
@@ -0,0 +1,30 @@
...
+A DMX universe contains 512 channels stored as bytes. This is too large to expose as a single register. 
+Are pipes the way to go?
+
+## Registers
+
+    rw enabled?: bool @ 0x80
+
+Determines if the DMX bridge is active
+
+    rw terminate?: bool @ 0x80
```
mmoskal, line 17: I don't think the controller (DMX master) ever terminates; there could a terminate switch on some of the slaves, but they would not talk JACDAC likely

services/dmx.md
mmoskal: (we don't specify format in modelrunner, iothub, tcp)

pelikhan:
I'm not quite sure how to handle incoming packets. Maybe we can support packets < ``n`` bytes on the incoming channel so that we can fit them in a single jacdac packet and stuff them into an event. ``n`` to be decided.

mmoskal:
I think we don't need incoming packets here.

The send command returning pipe is a bit wasteful - for a 300 byte packet it requires:
* send command
* pipe port report
* first data packet
* ack for it
* second data packet (with close flag)
* ack for it

I guess the ideal case would be:
* send command and first half of data
* "continuation" command with second half of data
* ack for both

I guess while, it looks like 2x overhead (3 packets vs 6), the actual time on the line is only ~20% less for the optimal option (every packet has around 360us overhead).

For a 50 byte packet it grows to ~40%.

Needs some thought.

cc @jamesadevine

pelikhan:
Or support smaller payloads. Which is ok for tinkering.

pelikhan:
Another option is to support 236 channels; which for tinkering is plenty. Articulated lights typically may use 8/40 of them.

mmoskal:
I agree, for this service using 236 channels should be enough

PR #203, opened 2021-01-26
Service: 7-segment display

pelikhan:
These displays are notorisouly annoying to wire up, but generally very useful.


services/7segmentdisplay.md
```diff
@@ -0,0 +1,33 @@
...
+
+The digit and dots displayed on the segment. If ``digit`` is outside ``0x0-0xf``, nothing is displayed.
+
+    rw brightness: u8 @ intensity
+
+Controls the brightness of the LEDs. ``0`` means off.
+
+    rw column: bool @ 0x80
+
+Show or hide the column. If the column is not supported, the value remains false.
```
mmoskal, line 23: what's column?

services/7segmentdisplay.md
```diff
@@ -0,0 +1,33 @@
...
+
+    const digit_count? = 4: u8 @ 0x80
+
+The number of digits available on the display.
+
+## Commands
+
+    command clear @ 0x80 {}
```
mmoskal, line 31: I think it's redundant with the value register

services/7segmentdisplay.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Registers
+
+    rw digits @ value {
+    repeats:
+        digit: u8
+        dot: bool
+    }
+
+The digit and dots displayed on the segment. If ``digit`` is outside ``0x0-0xf``, nothing is displayed.
```
mmoskal, line 15: I think we should assign each bit to a position in the display. We can go A to bit 0, B to 1, ... G to 6, and dot to 7.

150px-7_Segment_Display_with_Labeled_Segments svg

This way you can encode the different ways of showing say the number 7, but you can also try to display some approximation of text. It would be up to client to keep a "font" - mapping from numbers (and possibly letters) to 8 bit patterns.

This also lets you clear the display. We should also say what happens if display has say 7 digits and you send 4 - I guess the last 3 should be set to 0.

services/7segmentdisplay.md
pelikhan: some segments have double dot LED in the middle, to display time.

services/7segmentdisplay.md
pelikhan: removed

services/7segmentdisplay.md
pelikhan: I changed to turn off the remaining leds?

services/7segmentdisplay.md
```diff
@@ -0,0 +1,35 @@
...
+ |   |   -
+ E   C  |DP|
+ - D -   -
+```
+
+    rw brightness: u8 @ intensity
+
+Controls the brightness of the LEDs. ``0`` means off.
+
+    rw column_leds?: bool @ 0x80
```
pelikhan, line 29: column_dots

pelikhan: double_dots

PR #210, opened 2021-01-28
Enfore use of u0.... for / units

pelikhan:
/ fractions should use a u0. decimal representation so that we don't have to apply transformation after packing.
- [x] encoding of i0.16 looks broken. I send 0.8 and get back -0.2 in return. It seems to rescale things in -0.5, 0.5. is that what we want? @mmoskal 
- [x] there is an annoying overflow issue with signed encodings for i0.16, 0.5 -> -0xfffe - it overflows. So if you take the slider all the way to the right and drops to 50%


services/ledpixel.md
```diff
@@ -67,12 +67,12 @@ number, and followed by 3 bytes of color.
 
 ## Registers
 
-    rw brightness = 15: u8 / @ intensity
+    rw brightness: u0.8 / @ intensity
```
mmoskal, line 70:
```suggestion
    rw brightness = 0.05: u0.8 / @ intensity
```

mmoskal (approved):

services/ledpixel.md
pelikhan: = 0.05 does not parse

mmoskal: I see. My TODO keeps growing: 

## Spectool
* check if events are under events in spectool
* check if typical max etc are in range in spectool
* `u8 /` to `u0.8 /` in spectool (error)
* check if something follows undefined-sized type (eg: repeats: x: string) in spectool
* const register at 0x80 doesn't cause compile error
* disallow i0.X (only i1.X is ok)
* = 0.05 (led pixel bright) does not parse

I guess you did one though :)

pelikhan: if you have a issue todo i can add notes :)

PR #211, opened 2021-01-29
Updated role manager

### LINKED ISSUE #130: role manager should bind to device + service index
This would allow binding to a service that is not unique on a device.

example: device has multiple buttons, need to be able to bind to specific buttons.

pelikhan:
fixes #130

### DIFF
--- a/services/rolemanager.md
+++ b/services/rolemanager.md
@@ -1,15 +1,25 @@
 # Role Manager
 
-    identifier: 0x1
+    identifier: 0x1e4b7e66
 
-Assign roles to devices on the JACDAC bus.
+Assign roles to services on the JACDAC bus.
+
+Internally, the role manager stores a mapping from from `(device_id, service_idx)` to role name.
+Users refer to services by using role names (eg., they instantiate an accelerometer client with a given role name).
+Each client has a role, and roles are unique to clients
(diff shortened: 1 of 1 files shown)


services/rolemanager.md
```diff
@@ -1,6 +1,6 @@
 # Role Manager
 
-    identifier: 0x119c3ad1
+    identifier: 0x1
```
pelikhan, line 3: typo?

services/rolemanager.md
```diff
@@ -9,21 +9,25 @@ Assign roles to devices on the JACDAC bus.
     ro all_roles_allocated: bool @ 0x181
 
 Indicates if all required roles have been allocated to devices.
+TODO is this needed?
```
pelikhan, line 12: this is a cheap way to ask the role manager some roles need to be allocated. Otherwise, one has to setup a pipe for required_roles, etc...
We could replace it with a status code in the status register.

services/rolemanager.md
mmoskal: kind of, I just want a new identifier here - that's a way for the spectool to suggest one

mmoskal:
@pelikhan does this look reasonable now?

pelikhan:
Looks good

PR #213, opened 2021-02-02
weight scale service

pelikhan:


services/weightscale.md
```diff
@@ -0,0 +1,43 @@
+# Weight Scale
+
+    identifier: 0x1f4d5040
+    extends: _sensor
+
+A weight measuring sensor.
+
+## Registers
+
+    ro weight: u16.16 g @ reading
```
mmoskal, line 10: This is up to 65kg, which is not good for human scales.

I think we should use u22.10 - up to 4 tons and with 0.001g precision

services/weightscale.md
```diff
@@ -0,0 +1,43 @@
...
+
+The reported weight.
+
+    ro weight_error?: u16.16 g @ reading_error
+
+The estimate error on the reported reading.
+
+    rw zero_offset: u16.16 g 0x80
+    
+Calibrated zero offset error on the scale, i.e. the measured weight when nothing is on the scale.
```
mmoskal, line 20:
```suggestion
Calibrated zero offset error on the scale, i.e. the measured weight when nothing is on the scale.
You do not need to subtract that from the reading, it has already been done.
```

services/weightscale.md
```diff
@@ -0,0 +1,43 @@
...
+
+    rw gain: u16.16g 0x81
+    
+Calibrated gain on the weight scale error.
```
mmoskal, line 24: Shouldn't this be some sort of multiplier?

services/weightscale.md
```diff
@@ -0,0 +1,43 @@
...
+
+    enum Variant: u32 {
+        LoadCell = 1
+    }
+    const variant?: Variant @ variant
```
mmoskal, line 29: BodyWeight, Kitchen, Jewelry ?

services/weightscale.md
pelikhan: Yes internally it's factor applied to the weight, offset + factor * measured_weight

mmoskal: The I think this should be unit-less, if it's a multiplier.

Also good to make these things optional - I imagine some chips may just store that stuff internally and not give you access.

mmoskal:
we should also have `max_weight` (which sort of overlaps with variant, but that's all right)

mmoskal (approved):

PR #214, opened 2021-02-02
Service: GPS

pelikhan:


services/gps.md
```diff
@@ -0,0 +1,58 @@
...
+    extends: _sensor
+    camel: gps
+
+A Global Positioning Device. NMEA message reference at https://www.sparkfun.com/datasheets/GPS/NMEA%20Reference%20Manual-Rev2.1-Dec07.pdf.
+
+## Registers
+
+    ro position @ reading {
+        timestamp: u32 utc_time
+        latitude: i12.20 ° { absolute_min=-90, absolute_max=90 }
```
mmoskal, line 13: i9.23 would be enough for both (lat only needs i8.24 but let's keep it consistent)

services/gps.md
```diff
@@ -0,0 +1,58 @@
...
+    identifier: 0x19dd6136
+    extends: _sensor
+    camel: gps
+
+A Global Positioning Device. NMEA message reference at https://www.sparkfun.com/datasheets/GPS/NMEA%20Reference%20Manual-Rev2.1-Dec07.pdf.
+
+## Registers
+
+    ro position @ reading {
+        timestamp: u32 utc_time
```
mmoskal, line 12: I think we should have millisecond-precision time here; maybe u64?

services/gps.md
```diff
@@ -0,0 +1,58 @@
...
+        latitude: i12.20 ° { absolute_min=-90, absolute_max=90 }
+        longitude: i12.20 ° { absolute_min=-180, absolute_max=180 }
+    }
+
+Reported position and time of position.
+
+    rw enabled?: bool @ intensity
```
mmoskal, line 19: shouldn't be optional; you definitely want to turn it off - it's super power-hungry

services/gps.md
```diff
@@ -0,0 +1,58 @@
+# GPS
+
+    identifier: 0x19dd6136
+    extends: _sensor
+    camel: gps
```
mmoskal, line 5: GPS is a particular American system. Should we have something more neutral and say which system we're using (GPS, GLONASS, Galileo, and the Chinese one). There are some chips that  can do several at once improving lock speed.

services/gps.md
pelikhan: The serial interface reports seconds.

services/gps.md
pelikhan: Navigation system?

services/gps.md
mmoskal: adafruit guide has this example `$GPRMC,194509.000,A,4042.6142,N,07400.4168,W,2.03,221.11,160412,,,A*77` where the time is `19:45:09.000`; it is `.000` but I think it could have more precision; GPS does have a very precise time

services/gps.md
mmoskal: positioning system; or maybe "satnav"?  (this is how the Brits call it)

services/navigationsystem.md
```diff
@@ -0,0 +1,54 @@
...
+    camel: satnav
+
+A satellite-based navigation system like GPS, Gallileo, ...
+
+## Registers
+
+    ro position @ reading {
+        timestamp: u64 ms
+        latitude: i9.23 ° { absolute_min=-90, absolute_max=90 }
+        longitude: i9.23 ° { absolute_min=-180, absolute_max=180 }
```
mmoskal, line 14: should we add error in lat/long separately?

pelikhan:
Yes I need to fix that.

pelikhan:
So there is still a question whether to expose altitude, course (direction) and speed on ground.

mmoskal:
altitude for sure; i'm not sure about the other two, do GPS units typically expose that?

JohnVidler:
Hey @pelikhan - we're looking to use jacdac with some educational kits as part of a research/accelerator project, and would need GPS services to do this.

What are the current blockers (if any) on this for merging, and can we help get this integrated in any way? We don't have real hardware to test this at the moment, but we're prepared to spin up some boards if this is whats needed here.

danielstl:
From what I could tell, it seems that the latitude/longitudes should have their datatypes updated (from degrees to the specific latitude & longitude types), and probably having the reading_error values extracted into their own registers?

pelikhan:
I've removed reading_error for now. Does the current service look like a good start for you?

pelikhan:
@danielstl  @JohnVidler

danielstl:
Yeah looks good so far - cheers!

PR #215, opened 2021-02-02
Service: RFID Reader

pelikhan:

Instead of keeping some kind of list of tags, I"m keeping this service lightweight and let the client catch the events.

There does not seem to be any upper bound on the size of a tag... so this won't work for long payloads.


mmoskal (approved):

services/rfidreader.md
```diff
@@ -0,0 +1,20 @@
+# RFID Reader
+
+    identifier: 0x1247aec2
+    extends: _sensor
```
mmoskal, line 4: should we extend sensor here?

pelikhan: huh not anymore.

pelikhan:
this one needs more taughts. What kind of API do we want?

mmoskal:
I don't think I know enough about RFID to comment...

PR #216, opened 2021-02-03
Service: bit:radio

pelikhan:


services/nrfradio.md
```diff
@@ -0,0 +1,36 @@
...
+Group used to filter packets
+
+    rw transmission_power = 6: u8 { min_value=1, max_value=7 } @ 0x81
+
+Antenna power to increase or decrease range.
+
+## Commands
+
+    command send @ 0x80 {
+        data: bytes { max_bytes: 18 }
```
mmoskal, line 22: I don't think we should limit this to what micro:bit v1 can do; the max len of packets is in the ~250 bytes range.

There is some `S0` and `S1` fields that I don't understand. Also I think that `group` is a codal concept (maybe it's just the first byte of payload?).

@jamesadevine should be able to say more.

services/nrfradio.md
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+Sends a payload of bytes as a radio message
+
+## Events
+
+    event packet @ 0x80 {
+        time: u32 ms
+        rssi: i16 db
+        device_serial_number: u32
```
mmoskal, line 32: I think the device serial is m:b-specific; in fact there seem to be some addresses in the packets themselves, maybe these are MAC addresses of BLE? @jamesadevine ?

services/nrfradio.md
pelikhan: I was thinking of explicately support a v1-backward compatible radio protocol. Maybe we can have something more generic as another service.
The purpose of this service would be bridge JACDAC buses over radio too.

jamesadevine: The radio is restricted in software to communicate 32 bytes max (I think).

The max data length of a radio packet can be 255 bytes, not including metadata.

My big question is what is the purpose behind this service? bit:radio is a lossy protocol, it has no reliability guarantees and collisions between devices would be likely if running jacdac on top (if two devices communicate/overlap packets then both packets get lost).

If this service is simply to send and receive bit:radio packets with no reliability guarantees, then fine, great. But if you're envisaging jacdac over radio it would require a mesh protocol with TDMA management (what I did as part of my thesis) for reliability.

services/nrfradio.md
jamesadevine: AFAIK, micro:bit does not include the mac address of the hardware to make PII data in packets hard to trace back to a source. The group id (address) provides a group abstraction for user applications. If some sort of identifier is required it is consciously added by the user. The radio hardware can filter based on the given group id (address).

See: https://github.com/lancaster-university/microbit-dal/blob/master/inc/drivers/MicroBitRadio.h#L80

services/nrfradio.md
pelikhan: > If this service is simply to send and receive bit:radio packets with no reliability guarantees, then fine, great.
Yes that's the goal. Imagine a cheap module with nrf51 that exposes sensor via radio. It's another way to look at JACDAC.

services/nrfradio.md
pelikhan: I'm renaming this to "bit:radio". this is specifically to communicate with other microbits.

services/nrfradio.md
jamesadevine: I think a basic comms channel between bit:radio and jacdac sounds good. To run jacdac on a properietry radio, we would have to do something radically different (i.e. think about wireless in the same way as we did with single wire serial).

Remember packet payload is capped at 32 bytes!! :smile:

pelikhan: Not even that. We reserve some bytes in MakeCode for time/serial etc...

mmoskal: should we replicate makecode restrictions in this service?

pelikhan: Yeah, hence the 18bytes. Maybe we need the other variants, send string, send number, ...

services/nrfradio.md
```diff
@@ -0,0 +1,69 @@
...
+## Commands
+
+    command send_string @ 0x80 {
+        message: string { max_length: 18 }
+    }
+
+Sends a string payload as a radio message
+
+    command send_number @ 0x80 {
+        value: f64
```
pelikhan, line 31: @mmoskal do we support doubles?

mmoskal: I see no problem with this, especially that the JD service only need to pass on the bits, no need to do any double computations.

PR #217, opened 2021-02-04
Service: Random Number Generator

pelikhan:


services/rng.md
```diff
@@ -0,0 +1,25 @@
+# Random Number Generator
+
+    identifier: 0x1789f0a2
+    camel: rng
+
+A service that generates true random number generated by a HRND or TRNG.
```
mmoskal, line 6:
```suggestion
A service that generates random numbers using, where entropy is sourced from physical processes.
```

services/rng.md
```diff
@@ -0,0 +1,25 @@
...
+## Commands
+
+    command random @ 0x80 {
+        length: u8 { absolute_max=128 }
+    }
+    report random {
+        data: bytes
+    }
+
+A command that generates a random buffer with the given length.
```
mmoskal, line 17:
```suggestion
A command that generates a random buffer with the given length.
The response to this command is delayed when there isn't enough entropy available (similar to `/dev/random`).
```

Also, let's make this one optional, and add another one, I would call it `urandom` that doesn't wait for enough entropy, just uses a crypto-secure PRNG, seeded with real entropy.

services/rng.md
```diff
@@ -0,0 +1,25 @@
...
+
+    enum Variant: u32 {
+         TRNG = 1
```
mmoskal, line 20:
```suggestion
         Hardware = 1
```

mmoskal:
also a register `entropy_available: u32 bytes` would be nice

mmoskal:
I think we should make a distinction like Linux - https://en.wikipedia.org/wiki//dev/random

mmoskal:
or maybe not, "In 2020, the Linux kernel version 5.6 /dev/random only blocks when the CPRNG hasn't initialized. Once initialized, /dev/random and /dev/urandom behave the same."

mmoskal (approved):
OK, forget entropy estimate, urandom, and blocking - this seems no longer used.

I suppose there are two possible hardware impl here. Both would use PRNG (eg Fortuna) to generate numbers, and they would re-seed periodically using either a real hardware generator made of transistors (which are noise due to quantum effects), or use the ADC noise from temperature reading, or some floating pin.

pelikhan:
@mmoskal i think i'll let you finalize this one.

mmoskal:
I guess it was already done when you asked :)

PR #218, opened 2021-02-04
MIDI output

pelikhan:


mmoskal:
I don't feel qualified to comment...

pelikhan:
This API is mostly compatible with WebMIDI and any output MIDI device. The buffer should contain midi data which is well documented and handled elsewhere.

mmoskal (approved):
lgtm

PR #219, opened 2021-02-04
Add hold to button spec

jamesadevine:


jamesadevine:
why do up/down appear at 0x1/0x2, but click, long click, and hold come at 0x8N?

pelikhan:
down/up map to reserved event code active/inactive

jamesadevine:
Got it... so presumably the hold button code needs to be a different value?

pelikhan:
Looks good

mmoskal (approved):

Issue #113, opened 2020-12-19
"pure/idempotent" commands


pelikhan:
I am proposing to annotate commands with a "pure" attribute, which means that the command can be called without changing the internal state of the device.

This is usefull, for example, to automatically display the result of "list required roles" in the role manager, without running the risk of launching a rocket.

mmoskal:
`list required roles` should logically be a read-only register, but the response may be too long, so it's a command opening a pipe. Maybe we should have syntax for pipe-registers?

pelikhan:
Yes pipe register sound good.

note: MakeCode piped time out silently and returns partial data. Needs fixing

mmoskal:
Never did this.

PR #223, opened 2021-02-06
Service: ePaper

pelikhan:

### DIFF
--- a/services/epaper.md
+++ b/services/epaper.md
@@ -1,0 +1,39 @@
+# ePaper
+
+A e-Ink/e-Paper display.
+
+    identifier: 0x17a19937
+
+## Registers
+
+    rw enabled: bool @ intensity
+
+Turns on/off the display.
+
+    const rows: u16 @ 0x80
+
+Number of rows on the screen.
+
+   const columns: u16 @ 0x81
+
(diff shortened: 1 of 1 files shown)


services/epaper.md
```diff
@@ -0,0 +1,39 @@
...
+    const variant: Variant @ variant
+
+The type of screen and number of bits per pixels. Monochrome = 1bit, Grayscale4 = 2bits, Tricolor = 2bits.
+
+## Commands
+
+    command paint @ 0x80 {
+        colors: pipe
+    }
+    pipe report colors {
```
pelikhan, line 35: this should be the other way around, it should be a pipe towards the display

mmoskal: yeah, this is one of these "big registers"  - #113 

this is usually done like the `set_model` here
https://github.com/microsoft/jacdac/blob/main/services/modelrunner.md#commands

(you can skip the size, and you would have `pipe command colors { ... }`

pelikhan:
@mmoskal updated

mmoskal (commented):
looks good, but I think we need some more info on how the bits are packed (row-major or column-major? are rows/columns aligned to byte/half word/word?)

PR #224, opened 2021-02-06
simplify led

pelikhan:
Allow to "just" set the color on an LED without going through animations
+ specify repetitions in animation format.


services/led.md
```diff
@@ -7,12 +7,21 @@ A controller for 1 or more monochrome or RGB LEDs connected in parallel.
 
 ## Registers
 
+    rw color @ 0x80 {
+        red: u8
+        green: u8
+        blue: u8
+    }
+
+Sets or gets the color on the LED. Settings the color cancels any animation running on the light.
```
mmoskal, line 16: I'm not sure. It seems to have the same effect as a constant animation with that color. I feel this just complicates the service.

services/led.md
```diff
@@ -22,7 +31,7 @@ At `0` the power to the strip is completely shut down.
 
 Animations are described using pairs of color description and duration, 
 similarly to the `status_light` register in the control service.
-They repeat indefinitely until another animation is specified.
+`repetation` as 0 is considered infinite.
```
mmoskal, line 34: sounds good!

mmoskal (approved):

PR #225, opened 2021-02-07
Service: Compass

pelikhan:

### DIFF
--- a/services/compass.md
+++ b/services/compass.md
@@ -1,0 +1,22 @@
+# Compass
+
+A sensor that measures the heading.
+
+    identifier: 0x15b7b9bf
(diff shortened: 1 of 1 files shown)


services/compass.md
```diff
@@ -0,0 +1,22 @@
+# Compass
+
+A sensor that measures the heading.
+
+    identifier: 0x15b7b9bf
+    extends: _sensor
+
+## Registers
+
+    ro heading: u16 { absolute_min=0, absolute_max=359, preferred_interval=1000 } ° @ reading
```
mmoskal, line 10: I think this should be `u16.16` - there are surely compasses better than 1deg

PR #227, opened 2021-02-08
Service: Barcode reader

pelikhan:
note: dfrobot sells vision accessories that can detect barcodes.

### DIFF
--- a/services/barcodereader.md
+++ b/services/barcodereader.md
@@ -10,7 +10,7 @@ A device that reads various barcodes, like QR codes. For the web, see [BarcodeDe
     
 Turns on or off the detection of barcodes.
 
-    enum Format: u32 {
+    enum Format: u8 {
         Aztec = 1
         Code128 = 2
         Code39 = 3
--- a/services/barcodereader.md
+++ b/services/barcodereader.md
@@ -36,7 +36,7 @@ Reports the list of supported barcode formats, as documented in https://develope
 
     event detect @ active {
         format: Format
-        data: bytes 
+        data: string
     }
     
 Raised when a bar code is detected and decoded. If the reader detects multiple codes, it will issue multiple events.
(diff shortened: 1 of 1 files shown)


services/barcodereader.md
```diff
@@ -0,0 +1,43 @@
...
+
+A device that reads various barcodes, like QR codes. For the web, see [BarcodeDetector](https://developer.mozilla.org/en-US/docs/Web/API/BarcodeDetector).
+
+## Registers
+
+    rw enabled: bool @ intensity
+    
+Turns on or off the detection of barcodes.
+
+    enum Format: u32 {
```
mmoskal, line 13:
```suggestion
    enum Format: u8 {
```
Gives you three more bytes in the detect payload; some URLs are long

services/barcodereader.md
```diff
@@ -0,0 +1,43 @@
...
+            format: Format;  
+    }
+    
+Reports the list of supported barcode formats, as documented in https://developer.mozilla.org/en-US/docs/Web/API/Barcode_Detection_API.
+
+## Events
+
+    event detect @ active {
+        format: Format
+        data: bytes
```
mmoskal, line 39:
```suggestion
        data: string
```

the webapi uses strings, and I think none of these barcodes are binary
it also makes it clear that EAN13 should read as `"088381463935"` and not say bytes `0x00, 0x08, 0x08, ...`

services/barcodereader.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Events
+
+    event detect @ active {
+        format: Format
+        data: bytes 
+    }
+    
+Raised when a bar code is detected and decoded. If the reader detects multiple codes, it will issue multiple events.
+
```
mmoskal, line 43:
```suggestion
In case of numeric barcodes, the `data` field should contain the ASCII (which is the same as UTF8 in that case) representation of the number.
```

mmoskal (approved):

services/barcodereader.md
pelikhan: Good catch. Web api is strong too

PR #258, opened 2021-02-23
use register for RNG

pelikhan:
Use register to get random bytes on every read.


services/rng.md
```diff
@@ -10,15 +10,10 @@ which is periodically re-seeded with entropy coming from some hardware source.
...
-        length: u8 { absolute_max=128 }
-    }
-    report random {
-        data: bytes
-    }
+    ro random: bytes @ reading
 
-A command that generates a random buffer with the given length.
-This never blocks for a long time.
+A register that returns a 128 bytes random buffer on every request.
```
mmoskal, line 15: I would go with 64. People rarely need more than 16 bytes (UUID) or maybe 32 (AES 256). Packet overhead is 32 bytes is it's doesn't make much sense to make it less than 32. So either 32 or 64 I would say.

pelikhan: 64 it is.

PR #262, opened 2021-02-26
Update test parser

tballmsft:
allow for javascript expressions and do code generation as well as checking of syntax.


dist/tests/_base.ts
```diff
@@ -0,0 +1,33 @@
+
+    import { JDService } from "../../../src/jdom/service"
+    import { JDServiceClient } from "../../../src/jdom/serviceclient"
+    import { JDRegister } from "../../../src/jdom/register"
+    import { useRegisterUnpackedValue } from "../../../docs/src/jacdac/useRegisterValue";
+    import { BaseReg } from "../specconstants";
+
+    namespace tests {
```
pelikhan, line 8: don't use namespace since you're already using modules

dist/tests/_base.ts
```diff
@@ -0,0 +1,33 @@
...
+            private reg_status_code: JDRegister
+            private status_code: number;
+                        
+  
+        
+            constructor(service: JDService) {
+                super(service);
+
+                this.reg_instance_name = service.register(BaseReg.InstanceName);
+                const [tmp_instance_name] = useRegisterUnpackedValue<[number]>(this.reg_instance_name);
```
pelikhan, line 23: hooks can only be used within the context of React function, I don't think that's what you what to do.

dist/tests/_base.ts
```diff
@@ -0,0 +1,33 @@
...
+    import { JDService } from "../../../src/jdom/service"
+    import { JDServiceClient } from "../../../src/jdom/serviceclient"
+    import { JDRegister } from "../../../src/jdom/register"
+    import { useRegisterUnpackedValue } from "../../../docs/src/jacdac/useRegisterValue";
+    import { BaseReg } from "../specconstants";
+
+    namespace tests {
+        export class BaseTestClient extends JDServiceClient {
+
+            private reg_instance_name: JDRegister
```
pelikhan, line 11: camelize

spectool/jdspectool.ts
```diff
@@ -70,14 +70,108 @@ function tsify(name: string) {
...
+        return (token.js ? token.js : "") + (token?.const ? token.const.toString() : "") + (token.id ? token.id : "");
+    }
+}
+
+// create a test monitor as a helper for manual tests
+
+// use reg.subscribe(CHANGE, () =>. ....) to listen for changes
+// await regchange(reg, v => v > 5), but need a hook for cancelling
+
+function toTestMonitor(spec: jdspec.ServiceSpec, serviceTest: jdtest.ServiceTest) {
```
pelikhan, line 120: move to jdtest

spectool/jdspectool.ts
```diff
@@ -70,14 +70,108 @@ function tsify(name: string) {
...
+                locals.push(l);
+        })
+    })
+    return locals;
+}
+
+// create a file to check syntax of expressions, if this file passes tsc
+// then toTestMonitorClient code will be fine
+
+function testSyntax(spec: jdspec.ServiceSpec, serviceTest: jdtest.ServiceTest) {
```
pelikhan, line 92: move to jdtest

pelikhan (changes requested):
I don't think we need test generation. what are the benefit w.r.t. to an interpreter of your JSON test structure + JDService/JDRegister?
Debugging generated tests is much more tedious because the dev cycle is longer (update generated, generate tests, debug generated tests)

tballmsft:
> I don't think we need test generation. what are the benefit w.r.t. to an interpreter of your JSON test structure + JDService/JDRegister?

It's useful for having the a full JavaScript expression as a conditional. I don't want to rewrite an expression interpreter.

pelikhan:
I get that but you’ll have to compile those and somehow load them dynamically. It’s a lot of boiler plate code.

tballmsft:
It's just a matter of precompiling and making a list of the tests with a uniform way to access them. There will be a single static import in the end, like all the other things we code generate.

tballmsft:
Maybe there's a way to generate json of expression from TypeScript AST and then just interpret that instead - would give more flexibility.

mmoskal:
Couldn't you just eval the conditional expressions (inside of interpreter). I think it would be safe to assume expressions are in JS not in TS?

pelikhan:
I’d rather avoid typescript expression all together and instead have a dsl. If the test requires more advanced code, it should be in a separate typescript file, with tooling support etc...

mmoskal:
yeah, maybe DSL would be enough - with function calls, maybe you can skip operators to simplify parsing?

tballmsft:
yeah, I think a simple expression parser and interpreter will give a lot more flexibility.

pelikhan:
So you could have a set of "sentences" that express various common scenario like "observe that event happens" and you can just parse out the ``event`` and stuff that into a prebaked test. It will be easier to read in the test ui.

tballmsft:
This looks good: https://github.com/EricSmekens/jsep

pelikhan:
This PR was not ready for merge.

PR #274, opened 2021-03-04
support for specifiying a custom specification url

pelikhan:
Configuration services for hardware devices should be customizable.


mmoskal:
I think it would be more user-friendly (and extensible) to have device description JSON URL in the control service, and have list of specification URLs there

spectool/jdspec.d.ts
```diff
@@ -563,6 +563,28 @@ declare namespace jdspec {
          */
         repo: string
 
+        /**
+         * Optional map of service class to a URL pointing to a service specification markdown file
+         */
+        serviceSpecificationsUrls?: {
+            [serviceClass: number]: string
```
mmoskal, line 570: Would it be much worse to just have a list of URLs and you figure out the service class from the markdown? less redundancy, so less chances for failure

JSON doesn't really have a map from number so it will look like "449414863": "https://..", but then people may want to use "0x1ac986cf": "https://..."

pelikhan:
Yes list of URLs is better

spectool/jdspec.d.ts
pelikhan: devices are stored in json already. https://github.com/microsoft/jacdac/blob/main/devices/microsoft/research/jmaccv2.json

mmoskal (approved):

PR #277, opened 2021-03-08
update prompt to be a function

tballmsft:
add more checking
- [x] start only within check
- [x] no nested starts
- [x] check kinds of testFun arguments


pelikhan (approved):

spectool/jdtestfuns.ts
```diff
@@ -15,43 +15,43 @@ export type Commands =
...
         id: "decreasesBy",
         args: ["reg", "number"],
-        prompt: "did the value of register $1 decrease by $2?",
+        prompt: (args) => `did the value of register ${args[0]} decrease by ${args[1]}?`,
     },
     {
         id: "rangesFromUpTo",
         args: ["reg", "number", "number"],
-        prompt: "register $1 should range in value from $2 up to $3",
+        prompt: (args) => `register ${args[0]} should range in value from ${args[1]} up to ${args[2]}`,
```
mmoskal, line 46: Not sure if you guys discussed that, but the old version was easier to translate

I would have used `"register {1} should range in value from {2} to {3}"` so that later you can have something like `"register {1} should range in value from {2:hex} to {3:hex}"` or whatever

tballmsft: oh, I forgot about

tballmsft: I will change back

PR #281, opened 2021-03-11
remove JSON path dependency

tballmsft:


pelikhan:
Also remove from package.json

PR #290, opened 2021-03-18
add spec for temperature

tballmsft:


services/tests/thermometer.md
```diff
@@ -0,0 +1,14 @@
+# Thermometer tests
+
+## in range
+
+Check that thermometer temperature is in expected range
```
pelikhan, line 5: since we render markdown on the UI, you can also write
```
Check that thermometer ``temperature`` is in expected range
```

PR #292, opened 2021-03-18
button spec

pelikhan:
PR to track changes. Feel free to comment, edit. @tballmsft 

Here are the related PRs: https://github.com/microsoft/jacdac-c/pull/15, https://github.com/microsoft/jacdac-ts/pull/181


tballmsft (commented):
Here's my suggestion for hold_time. Also, we agreed to add time to up event. I will do that next.

services/button.md
```diff
@@ -23,9 +23,10 @@ The amount of time a button needs to be held for a `hold` event to be generated.
 
 Emitted when button goes from inactive (`pressed == 0`) to active.
 
-    event up @ inactive
+    event up { time: u32 ms } @ inactive
```
pelikhan, line 26: i would say that u16 is enough here. where do we want to cap the time? how much can padauk handle?

We also have the 8ms unit if needed.

services/button.md
```diff
@@ -13,24 +13,26 @@ A push-button, which returns to inactive position when not operated anymore.
...
 
 Emitted together with `up` when the press time was not longer than 500ms.
 
-    event long_click @ 0x81
-
-Emitted after button is held for 500ms. Long click events are followed by a separate up event.
+    event hold @ 0x81
 
-    event hold @ 0x82
+Emitted after the button is held for more than `hold_time' (default/minimum 500ms). Hold events are followed by a separate up event.
```
pelikhan, line 37: ``
(default/minimum 500ms)
``

that's defined in the ``hold_time`` spec so i woudl not repeat here.

services/button.md
```diff
@@ -13,24 +13,28 @@ A push-button, which returns to inactive position when not operated anymore.
 
 Indicates whether the button is currently active (pressed).
 
+    rw hold_time: u16 { absolute_min=500 } ms @ 0x80
```
mmoskal, line 16:
```suggestion
    rw hold_time = 1000: u16 { absolute_min=500 } ms @ 0x80
```

mmoskal: also the 500 absolute min may be a bit low

mmoskal:
I think we should use `us` not `ms` for all time units - I think Padauk will only keep time in us not in ms. @jamesadevine  ?

services/button.md
pelikhan: i would say at least 1500 for hold, which allows for longer holds.

mmoskal:
though I guess we could keep time on padauk in 4us units, and pretend a ms is 1024us?

pelikhan:
``o`` -> 1024 || ``ous`` -> ``1024us`` ?

services/button.md
mmoskal: sorry, actually I meant too restrictive; maybe 100ms

tballmsft:
so then we need u32

tballmsft:
OK. How does it look now?

mmoskal:
I think we're still not sure if we want us or ms...

tballmsft:
agreed. I would say we should differentiate sensors dependent on human response time from those that are not dependent on human. For the former, ms. For the latter, us.

jamesadevine:
>  I think Padauk will only keep time in us not in ms

Padauk records time passed by a 6 ms timer interrupt!

tballmsft:
I reverted to ms and changed to one threshold (per @mmoskal's suggestion).

tballmsft:
I am finished implementing spec, code, and tests. All good and ready to submit.

pelikhan:
Can you link all the PRs together? Pasting URLs in the description is enough

services/button.md
```diff
@@ -13,24 +13,27 @@ A push-button, which returns to inactive position when not operated anymore.
...
+records the amount of time between the down and up events.
 
     event click @ 0x80
 
-Emitted together with `up` when the press time was not longer than 500ms.
-
-    event long_click @ 0x81
-
-Emitted after button is held for 500ms. Long click events are followed by a separate up event.
+Emitted together with `up` when the press time is not longer than `click_hold_time`.
```
pelikhan, line 35: not longer -> less than?

pelikhan:
LGTM

tballmsft:
Here are the related PRs: https://github.com/microsoft/jacdac-c/pull/15, https://github.com/microsoft/jacdac-ts/pull/181

PR #338, opened 2021-04-07
don't deprecate idenfity

pelikhan:
My understand is that we keep this command.


mmoskal (approved):
let's keep it for now

Issue #162, opened 2021-01-19
mixin services


pelikhan:
These are services that augment existing module and are connected to "parent" service. For example,

- [ ] LED inside an Arcade button
- [ ] analog position reader in a analog servo
- [ ] position encoder in a motor
- [ ] matrix keypad + LED matrix

* A auxiliary service is defined in the sub range ``0x4...``
* Auxiliary services must be following the parent service immediately in the control service list.

```
control
servo
|-- analog reader service
arcade button
|-- led service
```

Other idea:

```
-- light service
-- matrix keypad
--- matrix keypad light service
```

```
[0] control
[1] matrix keypad light service --> 2 sub services following me
[2] matrix keypad
[3] light service
```

mmoskal:
* let's call them mixin services, see https://en.wikipedia.org/wiki/Mixin
* let's use `0x2....` range

pelikhan:
We are not doing this.

mmoskal:
Well, I think we are, just not quite yet.

tballmsft:
Documentation needed

pelikhan:
https://microsoft.github.io/jacdac-docs/reference/service-specification/#mixins

PR #357, opened 2021-04-27
analog button service

pelikhan:
Refactoring the button service to support both digital and analog buttons by converting the boolean ``pressed`` reading into "pressure" reading, u0.16

rationale:
- [ ] we want a unified set of events for buttons, whether they are digital/analog
- [ ] users should preferrably use events, which abstract away debouncing, less packets on the bus
- [ ] pressed register was exposing a raw sensor data, which may not completly sync up with the debounced events.
- [ ] pressure button is now a "configuration service" which allows to set the threshold but not read the sensor.
- [ ] capacitive button is now a "configuration server" which allows to set the thresholds, calibrate but not read the sensor.

@jamesadevine this is probably a breaking change for you.


services/capacitivebutton.md
```diff
@@ -0,0 +1,25 @@
...
+
+    identifier: 0x1865adc9
+    tags: 8bit
+    group: button
+
+A configuration service for a capacitive push-button.
+
+## Registers
+
+    rw down_threshold: u0.16 / @ inactive_threshold
```
mmoskal, line 11: so the idea for up/down is that if the current value is between them, then the button retains previous state? once it crosses into say active range, then after a few frames it becomes active, and we wait until it crosses back into inactive for a few frames to de-activates it?

jamesadevine:
@pelikhan The touch test module is just for testing capacitive touch, so no breaking changes for me.

Does the padauk button implementation need updating before I flash the MCUs @mmoskal ?

pelikhan:
Yes let’s sync

services/capacitivebutton.md
pelikhan: right, we only need a active threshold

pelikhan: CODAL uses a single threshold

pelikhan: https://github.com/lancaster-university/codal-core/blob/master/source/drivers/TouchButton.cpp

mmoskal:
should we use the 0x2... range for the "mixin" services here? https://github.com/microsoft/jacdac/issues/162

I think we only need to update the spectool to accept that range, otherwise there should be no changes needed.

pelikhan:
@mmoskal https://github.com/microsoft/jacdac/pull/357/commits/abd3ab7f74a042c8b59c6946207c9d13d4d10601

mmoskal (approved):

PR #360, opened 2021-04-29
re-work joystick to also support gamepad

mmoskal:
This includes button status in joystick. This has three advantages:
* if you only care about digital joystick, you can just listen to change events (analog joystick should also simulate digital) and save on streaming
* it adds support for joystick button(s) - the small joystick usually have one, but gaming joysticks have plenty
* it also allows to treat gamepad as a joystick (from user perspective if a program works with small joystick, it should also work with a gamepad or a big joystick, right?)

This doesn't support the more complex gamepads (XBox/PS style), only an old-school Nintendo ones like in MakeCode Arcade. We could support XBox etc by adding "repeats:" in front of joystick axis and a few more buttons (Left/RightTrigger, Left/RightBumper)


services/joystick.md
```diff
@@ -2,30 +2,58 @@
...
+        Right     = 0x0000_0004
+        Down      = 0x0000_0008
+        A         = 0x0000_0010
+        B         = 0x0000_0020
+        Menu      = 0x0000_0040
+        Select    = 0x0000_0080
+        Reset     = 0x0000_0100
+        Exit      = 0x0000_0200
+        X         = 0x0000_0400
+        Y         = 0x0000_0800
```
pelikhan, line 24: do you want to add a button for the joystick itself?

pelikhan (approved):

jamesadevine (approved):

jamesadevine:
Honestly, we need better service extensibility. We are going to only see more and more combinations like this.

mmoskal:
It's composability not extensibility. For extensibility I think we have the mixin/configuration services which is reasonable to me. Composability we don't have much...

jamesadevine:
Right! Composability!

PR #361, opened 2021-04-29
mark button as stable

pelikhan:
Button is ready to be stabilized.


tballmsft:
Looks good to me, but I think we still need to discuss synthetic events, as we have lost the nice 'click' event.

jamesadevine:
Agreed! Peli and I were just chatting about this...

Issue #293, opened 2021-03-18
support for number constants in specs


pelikhan:
We have a few constants here and there in our specs that are not captured and code generated. I'm proposing to add syntax for numerical constants.

```
    constant: magic= 0x42
```

mmoskal:
I think this is a good idea, why close?

mmoskal:
I would rather use `const magic = 0x42` in the code section though, to align with enums

pelikhan:
it would be good to prioritize this so that we can move all the magic constants of the protocol in the spec and code-generate them for any environment. @mmoskal

pelikhan:
Implemented as "define foo 0

PR #369, opened 2021-05-06
support for consts in spec

pelikhan:
https://github.com/microsoft/jacdac/issues/293
- [x] C output
- [x] TS output
- [x] docs


spectool/jdspec.ts
```diff
@@ -391,6 +392,10 @@ export function parseServiceSpecificationMarkdownToJSON(
                 case "report":
                 case "command":
                 case "const":
+                    // test for const foo = 0x12
+                    if (words.length === 4 && words[2] === "=") constant(words)
```
mmoskal, line 396: this will parse "report x = 12" the same as "const x = 12"

also, i'm not sure I like the same keyword being used for two different things...

spectool/jdspec.ts
```diff
@@ -1240,12 +1255,11 @@ function prettyUnit(u: jdspec.Unit): string {
 }
 
 function toPython(
-    info: jdspec.ServiceSpec,
-    flavor: "micropython" | "circuitpython"
+    info: jdspec.ServiceSpec
 ) {
     const r = [
         "# Autogenerated C header file for " + info.name,
```
mmoskal, line 1261: autogenerated Python file for?

spectool/jdspec.ts
pelikhan: now ``define x 12``

PR #384, opened 2021-05-13
Generate .gts not .g.ts to avoid confusing intellisense

mmoskal:


pelikhan (approved):

pelikhan:
Also update the pxt.json generation

mmoskal:
I think pxt.json is fine - we generate pxt.g.json only if nothing is there

PR #401, opened 2021-05-25
client/server flags

pelikhan:


mmoskal (commented):
not a great fan of `server` wording; I like `lowlevel` more

mmoskal:
also client/server wording suggests it either one or the other, while most of our regs/cmds are none.

Interestingly, it's possible to think of a low-level client thing (used to implement higher-level client things). Then we would have to say `client server ro foobar`...

pelikhan:
ok for lowlevel. how about "metal"?

mmoskal:
metal better than server, but I think lowlevel is way more descriptive

PR #408, opened 2021-05-28
Service: HID Configurator

jamesadevine:


services/hidconfigurator.md
```diff
@@ -0,0 +1,53 @@
...
+        selector: u16
+        modifiers:u16
+    }
+Stores the given configuration on the server. If a configuration exists at this index, the new configuration will replace it.
+
+    command swap_to_configuration @0x82 {
+        configuration_number: u16
+    }
+    
+    command next_configuration @0x83 {
```
pelikhan, line 49: replace with current_configuration_index

services/hidconfigurator.md
```diff
@@ -0,0 +1,53 @@
...
+        service_class: u32
+        service_idx: u8
+        trigger_context:u8
+        trigger_value: u32
+        selector: u16
+        modifiers:u16
+    }
+Stores the given configuration on the server. If a configuration exists at this index, the new configuration will replace it.
+
+    command swap_to_configuration @0x82 {
```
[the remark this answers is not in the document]
jamesadevine: Sorry, these commands are for swapping between configurations

services/hidconfigurator.md
jamesadevine: I'll add add the current configuration register

services/hidconfigurator.md
```diff
@@ -0,0 +1,53 @@
...
+## Registers
+
+    rw num_configurations: u16 @ 0x80
+ 
+The number of configurations stored on this device.
+
+## Commands
+
+    command retrieve_configuration @ 0x80 {
+        configuration_number: u16
```
pelikhan, line 16: is number supposed to be unique identifier?

services/hidconfigurator.md
```diff
@@ -0,0 +1,53 @@
+# HID Configurator
+
+    identifier: 0x1e5758b5
+    
+A service for configuring how Jacdac device map to HID input events. Users can have multiple configurations and swap between them using the `swapTo` or `next` commands.
+
+## Registers
+
+    rw num_configurations: u16 @ 0x80
```
pelikhan, line 9: alternatively, we could have a command to return all configurations, and a clear command.

services/hidconfigurator.md
jamesadevine: Nope, just an index (i.e. 0 is the first one, 1 is the second one, etc.)

I'll add a comment to remove ambiguity

services/hidconfigurator.md
```diff
@@ -0,0 +1,53 @@
+# HID Configurator
```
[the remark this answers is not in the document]
pelikhan: see role manager for pattern

services/hidconfigurator.md
pelikhan: is u16 for alignment purposes?

jamesadevine: Not necessarily... u16 is a bit much... u8 would be fine!

services/hidconfigurator.md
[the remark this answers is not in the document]
jamesadevine: Neat idea!

pelikhan:
Also this is a configuration service for a "hid-hub" device or adapter. "hid-adapter" ?

services/hidconfigurator.md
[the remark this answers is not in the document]
pelikhan: you don't need the commands, you can set the ``current_configuration_index``

jamesadevine:
@pelikhan why does this not build? Padding?

mmoskal:
> @pelikhan why does this not build? Padding?

Yes

jamesadevine:
Also, how can I use the service editor after a PR has been made?

jamesadevine:
Is the format spec up to date? https://microsoft.github.io/jacdac-docs/reference/service-specification/#pack-format

mmoskal:
Pack format!= spec format. You need to use explicit reserved: u8/u16. In my experience register definitions say for mcu peripherals always have explicit padding like that.

jamesadevine:
But we're not defining MCU peripheral register maps, we are defining an abstract software interface?

mmoskal:
It's a concrete on-wire interface

jamesadevine:
But we do the work so that the engineer does not need to care about the padding. We generate the structs and you can only get so much information for debugging a wire. We have web based tooling that also removes the need to deal with this.

jamesadevine:
I'm not going to look at the specification for padding, I'm going to look at the h file...

mmoskal:
Well, if you're writing some wierd assembly then you may look at the spec. Also explicit padding makes you aware that you're possibly wasting bytes.

jamesadevine:
Do we have a doc that tells someone how to pad correctly? and/or can we have better output from the checker to tell us what needs to change??

mmoskal:
@jamesadevine I improved the padding message - it should be clear now; you need to decide if you want to re-order the fields, or if you want to add explicit padding (which is why this message is there in the first place)

pelikhan (approved):

services/hidconfigurator.md
```diff
@@ -0,0 +1,66 @@
+# HID Adapter Configurator
```
pelikhan, line 1: I would just call this "HID Adapter", the fact that the service allows configuration is a side-effect

pelikhan: and rename file to ``hidadapter``

jamesadevine: Done! Thanks @pelikhan

services/hidadapter.md
```diff
@@ -0,0 +1,66 @@
+# HID Adapter
+
+    identifier: 0x1e5758b5
+    
+A service for configuring how Jacdac device map to HID input events. Users can have multiple configurations and swap between them using the `swapTo` or `next` commands.
```
mmoskal, line 5: what commands?

services/hidadapter.md
```diff
@@ -0,0 +1,66 @@
...
+The current configuration the server is using.
+
+## Commands
+
+    command retrieve_configuration @ 0x80 {
+        configuration_number: u8
+    }
+    report {
+        configuration_number: u8
+        padding:u8[3]
```
mmoskal, line 24:
```suggestion
        padding: u8[3]
```

services/hidadapter.md
```diff
@@ -0,0 +1,66 @@
...
+    repeats:
+        device_id: u8[8]
+        service_class: u32
+        service_idx: u8
+        trigger_context:u8
```
mmoskal, line 29:
```suggestion
        trigger_context: u8
```

services/hidadapter.md
```diff
@@ -0,0 +1,66 @@
...
+        padding: u8[2]
+        trigger_value: u32
+        selector: u16
+        modifiers:u16
```
mmoskal, line 33:
```suggestion
        modifiers: u16
```

services/hidadapter.md
```diff
@@ -0,0 +1,66 @@
...
+    }
+    
+Retrieves a configuration stored on the server. If the configuration does not exist, an empty report will be returned
+
+    command store_configuration @ 0x81 {
+    configuration_number: u8
```
mmoskal, line 39:
```suggestion
        configuration_number: u8
```

services/hidadapter.md
```diff
@@ -0,0 +1,66 @@
...
+    padding:u8[3]
```
mmoskal, line 40:
```suggestion
        padding: u8[3]
```

services/hidadapter.md
```diff
@@ -0,0 +1,66 @@
...
+        service_idx: u8
+        trigger_context:u8
+        padding: u8[2]
+        trigger_value: u32
+        selector: u16
+        modifiers:u16
+    }
+Stores the given configuration on the server. If a configuration exists at this index, the new configuration will replace it.
+
+    command clear @ 0x82 {
```
mmoskal, line 53:
```suggestion
    command clear @ 0x82 { }
```

services/hidadapter.md
```diff
@@ -0,0 +1,66 @@
...
+    }
```
mmoskal, line 54:
```suggestion

```

services/hidadapter.md
```diff
@@ -0,0 +1,66 @@
...
+Clears all configurations stored on the device.
+
+## Events
+
+    event swapped @ change { }
+    
+Event that notifies clients that the server has swapped to a new configuration.
```
mmoskal, line 61: See comment on `modified`.

If you leave it, you could add current config index as an argument.

services/hidadapter.md
```diff
@@ -0,0 +1,66 @@
...
+
+    event modified @ 0x80 { }
+    
+Event that notifies clients that the server has modified to a configuration.
```
mmoskal, line 65: Does this trigger only when a `store_configuration` is received? Anyone interested can listen to that (and most likely the brain already knows it has modified the service). What's the use case scenario? We generally don't have events being triggered in response to explicit modifications.

mmoskal (changes requested):
Is the service intentionally limited to ~9 key bindings per config?

services/hidadapter.md
jamesadevine: I think I agree! Might be the result of some previous discussion... @pelikhan ?

services/hidadapter.md
jamesadevine: removed!

mmoskal: I can still see references to `swapTo` which doesn't exist?

services/hidadapter.md
```diff
@@ -0,0 +1,87 @@
...
+## Commands
+
+    command get_configuration @ 0x80 {
+        results: pipe
+        configuration_number: u8
+    }
+    pipe report configuration {
+        configuration_number: u8
+        binding_index: u8
+        device_id: u8[8]
```
mmoskal, line 26: we typically use `u64` (which requires 32 bit alignment) for device id

services/hidadapter.md
```diff
@@ -0,0 +1,87 @@
...
+        service_idx: u8
+        trigger_context:u8
+        trigger_value: u32
+        selector: u16
+        modifiers:u16
+    }
+    
+Retrieves a configuration stored on the server. If the configuration does not exist, an empty report will be returned
+
+    command set_configuration @ 0x80 {
```
mmoskal, line 37: this requires that there is a report from set_configuration that gives pipe_port, and then the client writes to that port, see https://github.com/microsoft/jacdac/blob/main/services/modelrunner.md#commands

services/hidadapter.md
```diff
@@ -0,0 +1,87 @@
...
+        bindings: pipe
+        configuration_number: u8
+    }
+    pipe report configuration {
+        configuration_number: u8
+        binding_index: u8
+        device_id: u8[8]
+        service_class: u32
+        service_idx: u8
+        trigger_context:u8
```
mmoskal, line 47: missing whitespace

services/hidadapter.md
```diff
@@ -0,0 +1,87 @@
...
+
+    command clear @ 0x82 {
+    }
+Clears all configurations and bindings stored on the device.
+
+## Events
+
+    event changed @ change { }
+    
+Event that notifies clients that the server has swapped to a new configuration or changed key bindings.
```
mmoskal, line 87: again, is this always due to commands being sent?

mmoskal (changes requested):
it seems that you committed my suggestions and they were lost then?

jamesadevine:
@mmoskal darn! This is what I get for using different machines at work and at home! I'll fix up, sorry

services/hidadapter.md
jamesadevine: @pelikhan mentioned emitting a changed event is better than guessing when the service is ready or polling the service until it is ready.

mmoskal: what do you mean "ready"? Does it take more than 1ms? We do not emit change event from a servo when you change the angle. cc @pelikhan

pelikhan: The rationale is that i query and cache the bindings in the client side. It's expensive and I don't want to do that constantly. servo angle is cheap actually.

services/hidadapter.md
```diff
@@ -0,0 +1,89 @@
...
+
+The current configuration the server is using.
+
+## Commands
+
+    command get_configuration @ 0x80 {
+        results: pipe
+        configuration_number: u8
+    }
+    pipe report configuration {
```
jamesadevine, line 23: @mmoskal is there a way I can reuse this report definition? we use it for both get and set!

services/hidadapter.md
jamesadevine: Updated

services/hidadapter.md
mmoskal: unfortunately no

services/hidadapter.md
```diff
@@ -0,0 +1,80 @@
...
+
+    command set_configuration @ 0x81 {
+        configuration_size: u32
+        configuration_number: u8
+    }
+    report {
+        configuration_port: pipe_port
+    }
+
+Open pipe for streaming a configuration to the service. The size of the configuration has to be declared upfront. The configuration is streamed over regular pipe data packets. When the pipe is closed, the configuration is written to flash.
```
mmoskal, line 46: what's the unit for size? (I suspect number of keys)

services/hidadapter.md
```diff
@@ -0,0 +1,80 @@
...
+    
+Retrieves a configuration stored on the server. If the configuration does not exist, an empty report will be returned
+
+    command set_configuration @ 0x81 {
+        configuration_size: u32
+        configuration_number: u8
+    }
+    report {
+        configuration_port: pipe_port
+    }
```
mmoskal, line 44: missing pipe command (repeated...) here

services/hidadapter.md
mmoskal: This needs some clear design guidelines. In particular, if no-one is talking to the servo, we should be able to assume its state did not change. If someone is, we should re-query (de-bounced probably). I think it would be good to assume that registers do not change, unless someone is sending commands to device, unless they have some modifier, say `volatile`. Services with `volatile` registers should then use the change event.

services/hidadapter.md
jamesadevine: @mmoskal Could we chat about this today? I'm not sure what I'm supposed to do here. Define a format for the pipe data?

mmoskal (approved):

services/hidadapter.md
```diff
@@ -0,0 +1,70 @@
...
+    pipe report configuration {
+        configuration_number: u8
+        binding_index: u8
+        padding: u8[2]
+        device_id: u64
+        service_class: u32
+        trigger_value: u32
+        trigger_context: u8
+        service_idx: u8
```
pelikhan, line 31: use full words -> service_index

services/hidadapter.md
```diff
@@ -0,0 +1,70 @@
...
+
+    command set_binding @ 0x82 {
+        configuration_number: u8
+        binding_index: u8
+        padding: u8[2]
+        device_id: u64
+        service_class: u32
+        trigger_value: u32
+        trigger_context: u8
+        service_idx: u8
```
pelikhan, line 46: same, use full words

services/hidadapter.md
pelikhan: Yes we should specify this. In this service or things like settings, it's costly to query pipes over and over again so I am trying to minimize those queries.

PR #448, opened 2021-06-07
Add orientation information to accelerometer spec

jamesadevine:


services/accelerometer.md
```diff
@@ -7,6 +7,18 @@
...
+
+An accelerometer module should translate acceleration values as follows:
+
+| Orientation           	| X value 	| Y value  	| Z value 	|
+|-----------------------	|---------	|----------	|---------	|
+| Module lying flat     	| 0       	| 0        	| -1000   	|
+| Module on "left" edge   	| -1000   	| 0        	| 0       	|
+| Module on "bottom" edge 	| 0       	| 1000     	| 0       	|
+
+We recomment an orientation marking on the PCB so that users can mount modules without having to experiment with the device.
```
pelikhan, line 20: this would be a brilliant example of test.

jamesadevine: Agreed! (cc @tballmsft)

services/accelerometer.md
```diff
@@ -7,6 +7,18 @@
...
+
+An accelerometer module should translate acceleration values as follows:
+
+| Orientation           	| X value 	| Y value  	| Z value 	|
+|-----------------------	|---------	|----------	|---------	|
+| Module lying flat     	| 0       	| 0        	| -1000   	|
+| Module on "left" edge   	| -1000   	| 0        	| 0       	|
+| Module on "bottom" edge 	| 0       	| 1000     	| 0       	|
+
+We recommend an orientation marking on the PCB so that users can mount modules without having to experiment with the device.
```
mmoskal, line 20: should be -1/1 not -1000/1000 (unit is g)

I would also say "module standing upright on "left" edge" to be more precise?

Also some wording like: "left/bottom can be determined by assuming text on silk runs left-to-right"

jamesadevine: Thanks! will edit

pelikhan (approved):
evenutally tests for this would be gold.

mmoskal (approved):
don't forget to run make and check in before merging

PR #479, opened 2021-06-15
Create qr-url-device-map.csv

jamesadevine:
Initial placement of QR code/url to device revision map.


devices/microsoft/research/qr-url-device-map.csv
```diff
@@ -0,0 +1,27 @@
+vanity name, design id, revision, module name, redirect source
```
pelikhan, line 1: did you reserve those? IMO, we shoud start with "jd*"

jamesadevine: They are all reserved.

jamesadevine: Feel free to go reserve some more! The assigned ones have shipped

pelikhan (approved):

PR #490, opened 2021-06-16
Add slider to device map

ducky64:
Adds the slider to the device map so I can give it a QR code. I don't know if the link is active, but if not someone should make it work?

Not yet at version 1.0, but pretty close and (probably) will be 1.0 when it gets sent to fab.


pelikhan (approved):

pelikhan:
You can also register your device here (another tool, we need to unify both). https://microsoft.github.io/jacdac-docs/tools/device-registration/

pelikhan:
The link will active with the build.

PR #497, opened 2021-06-27
update resolution

tballmsft:


spectool/jdutils.ts
```diff
@@ -66,7 +67,7 @@ export class SpecSymbolResolver {
 
     constructor(
         private readonly spec: jdspec.ServiceSpec,
-        private readonly role2spec: (role: string) => jdspec.ServiceSpec,
+        private readonly role2spec: (role: string) => { spec: jdspec.ServiceSpec, client: boolean},
```
pelikhan, line 70: run prettier.

pelikhan, line 70: considering client is the default scenario, refactor ``client: boolean`` to ``server?: boolean`` so that missing/falsy value means client.

PR #518, opened 2021-07-08
Button nanoservices

ducky64:
Introduces a set of button services (alongside the current button service) that takes more of a 'nanoservices' approach (see microsoft/jacdac#495) - separating the streaming button, edge (up/down) detection, and gesture (click, click-and-hold, double-click, ...) detection.

Implementation for the adapters (in JDOM) with a WIP unit test framework are in microsoft/jacdac-ts#219. That PR requires this one (with `yarn buildspecs`) for the service definition constants.

There's probably a larger question of what others issues exist that prevent this from working in practice. For example, needing VM support so the adapter code can run on any supported platform, how to expose multiple services (streaming and edge detection at least?) without running out of code space on the really, _really_ cheap micros, and whether there should be some notion of a stack of services or mutual exclusivity (can a service interface be claimed by another service?).


tballmsft:
I like the layering (dependence structure) of ro register -> low level events -> high level events. It's a nice way to decompose things, but maintaining nano-services has some overhead (separate files, many more services) that impacts tooling, search, stc.. Here are some thoughts:

1. edge detector needs to inherit from button, not sensor, correct? and so on...  If we were to have nano-services, it would be a must that the inheritance hierarchy allows us to find all the subclass of the low-level button service.

2. we could think of having "addressable" layers of a single service in which we would define events in layer N that depend on layers <N. These layers could form a namespace. The simplest button would only implement layer 0 (ro pressed) and probably layer 1 (edge detection). The implementation of layer 2 (higher level events) could be implemented on 

With (2), we could define all three nano-services (which I called layers above) in a single button service file , but still be able to refer to them (button.base, button.edge, button.high).

At a high-level, I think it's very useful to have a way to separate events into layers (nano-services) to make explicit the dependences. It also gives us a way to have a "minimal required" set of layers, with additional layers that can be implemented on server or client.

ducky64:
So one thing I'm trying to avoid with the nanoservices approach is having inflexible stacks. For example, while streaming button -> edge detect -> gesture makes sense, it might also make sense to layer a capacitive sensor below the streaming button, or have the capacitive sensor expose edges directly (with hysteresis or some fancier algorithm). I don't think the extends allows for that?

Some ideas:
- Perhaps we have the typical inheritance relation for the most common stack (here, the streaming button -> edge detect -> gestures). Modules could implement as high up the stack as they'd like, but must implement all the lower layers (so there's still some inflexibility here - unless we use non-conventional inheritance semantics).
  - Adapters must also implement all the lowest layers, though they could only implement the lowest layer and the upper layer adapters can handle the upper layer translation.
- But if we're taking a somewhat fresh look at design (and without thinking too hard about implementation difficulty, yet), the approach I'd lean towards is:
  - The nanoservices exposed to the user should be as nano as possible, but tooling and interfaces can help the user find the higher-level services (such as reachability analysis through adapters, or through some curated common adapters)
  - For efficiency reasons (code size for cheap micros, mainly?) it makes sense to have monolithic services. Two ideas here are:
    - Create a separate monolithic service that has adapters to the standard nanoservices. Tooling can make this more or less invisible (like the opposite of the `client` construct). In theory this should be a very low-cost (in terms of compute) abstraction (but the devil is going to be in the implementation details...)?
    - Alternatively, some concept of bundled services, either as syntactic sugar on top of the above, or some kind of service-union construct. Not sure what this would look like yet...

tballmsft:
We clearly can't afford to treat nano-services as separate Jacdac services (communicating via packets) when going to hardware. They would need to be combined into a monolith at compile time. Thus, my idea of layers. That is, you need a way to say "this button service is the composition of these nano-services, layered 1, 2, 3".

tballmsft:
@mmoskal and I chatted a little about this - one way forward is to have an "abstract" service that can be included by another service.  Such an abstract service can be reused by multiple services and present a common language of (presumably higher-level) events.  There's also a need to distinguish between registers/events/commands that are low-level and need to be implemented on the server, as well as those that are client side. The client side ones can be put into an abstract service. And, of course, an abstract service can include other (abstract) services. So this gives you the ability to define nano-services, but everything is combined together at compile time.

ducky64:
Yeah, I think that makes sense - though it also would seem the devil is in the details.

One concern about the composition of abstract services into one giant service is the possibility of address collisions. I don't think it's unsolvable, but it does require a bit of thought - whether the compiler does some remapping (perhaps addresses are specified with offsets and bounds?), or only certain services can be composed (might be very limiting?), 

Another thought might be to have separate constructs for these nanoservices from a user-facing view and machine-optimization view. Services currently fill both roles, and there's a bit of tension: smaller services (nanoservices) are more orthogonal (and I'd argue better APIs), while larger services (monolithic) are needed to optimize for cheap micros. I like the concept of interfaces (nanoservices) that can be compiled together into a machine-optimized definition (monolithic service) - though maybe there's also something to be said about being able to nest services arbitrarily deeply. That being said, it raises the question of when a device might expose multiple services, since they could always be compiled into a monolithic service.

Also, with nanoservices, does it make more sense to have the `client` tags associated with the nanoservice instead of individual events / registers? It seems the latter was because we had monolithic services that spanned levels of abstraction, but each nanoservice would (hopefully) only sit at one level of abstraction.

pelikhan:
Note that we already refer to "nano services" in our landing page so we need to find a different name for those.

ducky64:
Can we just go smaller - picoservices? 😁

ducky64:
Superseded by discussion microsoft/jacdac#567

PR #543, opened 2021-07-29
C# constant generation

pelikhan:
Generation C# classes for jacdac constants


pelikhan:
@fnawratil you can now use the generated .cs file in the dist folder

PR #570, opened 2021-08-24
add volatile as modifier on register

tballmsft:


tballmsft:
We've decided to add volatile as a modifier (to registers) as an optimization hint to clients for refreshing.

pelikhan (approved):

services/accelerometer.md
```diff
@@ -21,15 +21,15 @@ We recommend an orientation marking on the PCB so that users can mount modules w
 
 ## Registers
 
-    ro forces @ reading {
+    volatile ro forces @ reading {
```
pelikhan, line 24: aren't all readings implicitely volatile?

tballmsft: That's a good question. Are they?

tballmsft: Note I still need to go through all our specs to add volatile. If readings are implicitly volatile, it will certainly make this task easier.

pelikhan:
They are and there is already registers to specify the interval. volatile should only be allowed on rw registers

tballmsft:
> volatile should only be allowed on rw registers

I don't see why.

services/accelerometer.md
pelikhan: They are and there is already registers to specify the interval. volatile should only be allowed on rw registers

pelikhan: otherwise, it's a const

pelikhan:
``non-volatile read-only register`` == ``const``

tballmsft:
You shouldn't be able to "volatile const", agreed.

pelikhan:
>  A ro register can change value because you invoked a command.

Then, I don't understand the purpose of this flag. Could you add documentation for it in the spec mardown as well?

If a ``ro`` register is marked as non-volatile, then should the client refresh it?

tballmsft:
Yes, it is a hint for clients to do refreshing periodically.

pelikhan:
As specified this hint is not useful, since ``non-volatile ro`` can also be modified.

mmoskal:
The idea is that "non-volatile" registers only change because you invoked (non-get) command on the service (action or register write). Volatile registers can change for no reason obvious from looking at the bus. Thus for non-volatile you could have a caching strategy that flushes the cache whenever there is some write to the service and otherwise every 10 seconds or so (this is in case you drop some write packet). For volatile, you can cache say for 0.5s or so.

mmoskal:
So reading is a prototypical volatile register, but most other registers are non-volatile

tballmsft:
>     rw angle: i16.16 ° { typical_min=-90, typical_max=90 } @ value

I do have a question about angle of servo. Say it is at 30 degrees and then you write 90. It will take some time to move from 30 to 90. If you read the angle during this time, it may be some value between 30 and 90. Still, this is not volatile, because of the request to write 90, right?

mmoskal:
I think if you write 90 it means you want it to go to 90, and it will read as 90 immediately. There could be another register that gives you the current angle, and we could rename this one "requested_angle".

tballmsft:
Yes, that distinction (requested_angle, current_angle) makes sense.

Still my question stands, as the servo is in process of changing, is the current_angle register volatile?

pelikhan:
``current_angle`` is going to be volatile because an external force can be applied to the servo and change the angle. ``requested_angle`` is ``non-volatile``.

services/accelerometer.md
tballmsft: OK. Updated the code so that volatile only applies to ro/rw (and so comes after those).

services/_system.md
```diff
@@ -91,7 +91,7 @@ Asks device to stream a given number of samples
 
 Period between packets of data when streaming in milliseconds.
 
-    ro reading: i32 @ 0x101
+    ro volatile reading: i32 @ 0x101
```
pelikhan, line 94: reading is always volatile. i don't think we should require it.

services/servo.md
```diff
@@ -44,3 +44,8 @@ The servo motor will stop rotating when it is trying to move a ``stall_torque``
     const response_speed?: u16.16 s/60° @ 0x181
 
 Time to move 60°.
+
+    ro volatile current_angle?: i16.16 ° @ 0x182
```
pelikhan, line 48: this should be a reading register

spectool/jdspec.ts
```diff
@@ -571,6 +572,14 @@ export function parseServiceSpecificationMarkdownToJSON(
         if (words[0] === "internal") {
             internal = true
             words.shift()
+        } 
+        
+        let volatile: boolean = undefined
```
pelikhan, line 577: maybe automatically mark reading registers as volatile

services/_system.md
tballmsft: This is simply an annotation we provide in the system spec for information. We have the vocabulary "volatile" now - we should use it.  It's not a requirement - it's documentation.

services/servo.md
tballmsft: You can't have multiple reading registers in a spec. There's only one defined in _system:
 
> ro reading: i32 @ 0x101

tballmsft: Oh, I see: 

>     rw angle: i16.16 ° { typical_min=-90, typical_max=90 } @ value

is not reading.

pelikhan: Use reading instead of 0x182

spectool/jdspec.ts
tballmsft: that is done automatically by the system via inheritance.

services/_system.md
pelikhan: But all reading register will have to be marked with volatile?

spectool/jdspec.ts
tballmsft: my bad... need to do that.

tballmsft: Is it OK to have a service that has a reading register but doesn't extend _sensor?

pelikhan (approved):

spectool/jdspec.ts
pelikhan: yes, at least jacdac-docs ignore the sensor base class info since all packets are flattened when processing the specs.

spectool/jdspec.ts
```diff
@@ -650,6 +661,9 @@ export function parseServiceSpecificationMarkdownToJSON(
                 }
             }
 
+            if (packetInfo.identifierName === "reading")
```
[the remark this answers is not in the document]
tballmsft: right

pelikhan:
Should a modified non-volative register emit a change event?

spectool/jdspec.ts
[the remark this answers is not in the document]
mmoskal: Yeah, but reading is always ro.

Issue #571, opened 2021-08-25
reliable commands (dual of events)


tballmsft:
In analyzing our spec language and packets,  we see that events are the "reliable" version of reports.   So then, the question naturally arises, what's the "reliable" version of a command.  We are missing such a construct. It should be easy to implement, given the plumbing is already there for events.

mmoskal:
So the plan here would be to use the same `service_operation` range as events (`0x8000` to `0xffff`), but with the `command` flag set on the frame. This leaves us with a 7-bit counter for commands and 8-bit namespace for command codes. We should probably allocate a bit in the announce flags indicating this is supported, and then allow all commands to be sent via this channel. This (as well as the NACK feature) will definitely not fit on the smallest PADAUK, but I think we already decided to drop it in future.

(copied comment from #620 )

Issue #564, opened 2021-08-11
add command-not-understood report


mmoskal:
Right now we never know if the command (including register get/set) was lost or is it not supported. We could add a generic not-understood response.

One way to do it would be to use high bit of `service_number` field (and repeat `service_command` with `service_size==0`). I think that can be implemented in 1 bit of memory and 10-20 instructions on Padauk (which may still be too much for some services). This will be hacky, in the sense that if someone sends two commands very quickly one after another, and the first one is not understood, only the second one would be acted upon. The same is true for ACKs on Padauk though.

jamesadevine:
Is `service_number` == `service_index`?

According to the spec, this is what we use for ACKs:

<img width="1208" alt="Screenshot 2021-08-12 at 11 21 54" src="https://user-images.githubusercontent.com/4742697/129181071-74ccdc67-d413-468d-bb5f-6f2b38bb178b.png">

Are you suggesting we use one of these numbers (e.g. 0x3b) to represent a NACK, or something else?

pelikhan:
yes service_number -> service_index https://github.com/microsoft/jacdac-ts/commit/f27ae205d7b52c94527a5a7c10a7e1d3f4813fff

mmoskal:
I was thinking of doing `0x80 | service_index`

tballmsft:
makes sense to me - very useful.

pelikhan:
Much better than the retried we have to do now

pelikhan:
Updated at https://github.com/microsoft/jacdac/pull/809/files?authenticity_token=bd67Z9ReZKVxBJK9vV4QGIFr%2BhLAl41g74jmljXyFkSbvXUijk%2BaxjlooLOVqei5JFTDD%2BVIvlqv3Ohv3fWCGw%3D%3D&file-filters%5B%5D=.mdx#diff-228e4a143bb8029bfd3a1e61d49d59c2108ca871607774f5599b9c29b8d4dcd3

pelikhan:
The JDOM stores the information in the .notImplemented field of registers.

PR #809, opened 2021-10-14
Add reliable actions and error responses

mmoskal:
#571 #564


spec/reference/protocol.mdx
```diff
@@ -224,6 +224,27 @@ If these are handled purely through events, the client will maintain the pressed
...
+The `service_command` field is subdivided as follows:
+
+| Bits | Function 
+|------|----------------------
+| 15   | Always set for reliable actions
+| 14:8 | Sequential counter
+| 7:0  | Action code
+
+Only action codes in range `0x00-0xff` are supported.
+Just as for events, the sequential counter is per-device (but separate from the event sequential counter).
```
mmoskal, line 239: This doesn't work, since the client doesn't know the counter on the server. If client did know that, there could be collisions.

pelikhan:
Once we merge this i can start on the js side of things. Can i test against some existing firmware?

services/_system.md
```diff
@@ -48,6 +48,16 @@ Event from sensor or a broadcast service.
 
 Request to calibrate a sensor. The report indicates the calibration is done.
 
+    report command_not_implemented @ 0x03 {
```
pelikhan, line 51: "command" is a bit misleading here since it also applies to "register set/get"

mmoskal: register get/set is a command; the other thing is `action`; we should update our spec language, but the protocol docs are updated I think

PR #871, opened 2021-12-17
start on linking to hw and firmware

tballmsft:
We want to have device entry in catalog link to firmware source and hardware designs (as published)


pelikhan:
You need to update the DeviceSpec type in spec.d.ts

devices/microsoft/research/jmrotarycontrolbuttonv11.json
```diff
@@ -6,7 +6,9 @@
     "version": "1.1",
     "repo": "https://github.com/microsoft/jacdac-msr-modules",
     "link": "https://github.com/microsoft/jacdac-msr-modules",
+    "hardwareDesign": "https://github.com/microsoft/jacdac-ddk/tree/main/electronics/altium/reference-designs/JacdacRotary%2026-1.2",
```
pelikhan, line 9: Could we remove the space from these files to avoid having escaped space in URLs? (no more %20 please)

pelikhan:
I'll add the UI rendering once it's merged.

devices/microsoft/research/jmrotarycontrolbuttonv11.json
tballmsft: that will require a script as there are spaces everywhere  under the altium directory, and I am not sure that it wouldn't break altium unless we go into files and make sure any references to the files are renamed. Please file a separate issue.

pelikhan:
makecode-extensions.json should not have been modified by this change. Make sure to resync all submodules in jacdac-docs

PR #874, opened 2021-12-18
Service: Record and Replay

xmeow:


services/recordandreplay.md
```diff
@@ -0,0 +1,26 @@
...
+    
+    command record @ 0x81 {
+        seconds: u16
+    }
+    
+Record audio for N seconds.
+
+## Registers
+
+    rw seconds: u0.16 / @ intensity
```
pelikhan, line 23: Always less than a second? You probably want something like u22.10 s

services/recordandreplay.md
```diff
@@ -0,0 +1,26 @@
+# Record and Replay
+
+    identifier: 0x1b72bf50
+    group: sound
+    tags: 8bit, padauk
+    
+A record and replay module. You can record a few seconds of audio and play it back.
```
pelikhan, line 7: You probably want to document status code values so that you can use the status changed event to notify when the device is I. Recording/replay mode.

pelikhan, line 7: Volume control?

services/recordandreplay.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Commands
+
+    command replay @ 0x80 {}
```
pelikhan, line 11: Cancel command too?

services/recordandreplay.md
```diff
@@ -0,0 +1,26 @@
...
+    
+Replay cached audio.
+    
+    command record @ 0x81 {
+        seconds: u16
```
pelikhan, line 16: Maybe a register to document the maximum recording duration?

pelikhan (changes requested):

xmeow:
@pelikhan Please check the updated service define. Our current implementation is an MCU with only two gpio, one for record and one for replay. Cached time is handled by the protocol mcu.

services/recordandreplay.md
```diff
@@ -0,0 +1,44 @@
...
+
+    ro status: Status @ value
+
+Indicate the current status 
+
+    ro time: u16 ms / @ intensity
+    
+Milliseconds of audio recorded.
+
+    rw volumn?: u8 @ 0x180
```
pelikhan, line 41:
```
rw volume?: u8 @ intensity
```

services/recordandreplay.md
```diff
@@ -0,0 +1,44 @@
...
+        Idle = 0
+        Recording = 1
+        Playing = 2
+    }
+
+    ro status: Status @ value
+
+Indicate the current status 
+
+    ro time: u16 ms / @ intensity
```
pelikhan, line 37: ro time: u16 ms @ 0x180

services/recordandreplay.md
```diff
@@ -0,0 +1,44 @@
...
+    
+Replay cached audio.
+    
+    command record @ 0x81 {
+        milliseconds: u16 ms
+    }
+    
+Record audio for N milliseconds.
+
+    cancel record @ 0x82 {}
```
pelikhan, line 21:
```
command clear @ 0x82 {}
```
It clears the recorded soud right?

xmeow: No, just interrupt the current record session and keep the recorded samples. Our current implement is just a die chip with only 2 control io. But I think some advanced implement should be able to read/write/delete and list sound samples, stored in onboard spi-flash or SD card. Maybe I should check similar solutions in the market to make this service more complete~

pelikhan: oh great, let's do
```
command stop_recording @ 0x82 {}
```

pelikhan:
We have a "sound player" service defined here: https://microsoft.github.io/jacdac-docs/services/soundplayer . Basically, a list of sound names and a way to play/cancel them. Is there too much overlap with this service?

xmeow:
@pelikhan Sorry that we didn't aware of this service and revert the last commit, just keep things simple.

services/soundrecorderwithplayback.md
```diff
@@ -8,17 +8,17 @@ A record and replay module. You can record a few seconds of audio and play it ba
...
+Replay recorded audio.
     
     command record @ 0x81 {
         milliseconds: u16 ms
     }
     
 Record audio for N milliseconds.
 
-    cancel record @ 0x82 {}
+    command cancel @ 0x82 {}
```
pelikhan, line 21: @xmeow does cancel stops replaying or is it only for recording?

pelikhan:
@xmeow please take a look at the service to make sure it looks good for you

pelikhan (approved):

Issue #879, opened 2022-01-05
rename relay.closed -> relay.active in spec


mmoskal:
@sehodges says: you would normally think of 'actuating' a relay (rather than 'closing' it) because the majority of relays have changeover contacts (so in a sense they are always closed, whether actuated or not)

we should add some language that pins should be labelled: common, active, in-active or something?
(when "active=1" -> common is connected to active, and otherwise common is connected to in-active)

pelikhan:
are relay commonly name "normaly open" "normaly closed"?

mmoskal:
they typically have 3 terminals, so you chose if they are "normally open" (when you connect common and in-active) or "normally closed" (when you connect common and active).

pelikhan:
We used to have active/inactive and that confused users during the hackathon. 

https://github.com/microsoft/jacdac/commit/9d52682bff482c5bb503f617088c6e4f74e62bc9#diff-e9c53b9b51dd29173c184f63e7ec00b505b892bee0f5e21640f021ee69bd8cdf

mmoskal:
Did anyone get a rely module during the hackathon?

I think this can be solved with proper labelling, some subset of the following:

![image](https://user-images.githubusercontent.com/10673976/148365723-6f9cd899-08fe-413d-b6b3-7c0b8630f067.png)

mmoskal:
Or maybe like this:

![image](https://user-images.githubusercontent.com/10673976/148365950-7a0e8105-09a5-4e8d-9d77-c3bdb9d42200.png)

sehodges:
'A' would usually be 'NO' (for normally open) and 'I' would be 'NC' (for normally closed). 'C' is OK but could be 'COM' (for common) to disambiguate from the C in NC.

Steve

pelikhan:
maybe they got a switch. @jamesadevine  do you reckon?

pelikhan:
Should we align switch terminology as well?

mmoskal:
Jacdac switch doesn't have external contacts, so probably no need to change anything there.

PR #883, opened 2022-01-07
rename 'closed' to 'active' in relay + minor spec fixes

mmoskal:
See #879 

Also, `event` report is no longer used - removing.
Also change `auto_off_delay` encoding to be more reasonable (we don't need sub-ms precision there)

### DIFF
--- a/services/_system.md
+++ b/services/_system.md
@@ -36,13 +36,6 @@ The report format is the same as the format of the register.
 Registers number `N` is set by issuing command `0x2000 | N`, with the format
 the same as the format of the register.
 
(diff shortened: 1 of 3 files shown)


jamesadevine:
I think peli wanted to keep event report as it was/is an established pattern? Maybe I missed a conversation.

mmoskal:
We now have the reliable (triplicate) events, which occupy half the service command space. This wrapper report was no longer used anywhere. @pelikhan  ?

pelikhan (approved):

PR #890, opened 2022-01-17
update relative links so that they are correct

jamesadevine:


pelikhan (approved):
Best to use absolute links

PR #961, opened 2022-04-04
Air quality index

pelikhan:
See https://www.airnow.gov/?city=Seattle&state=WA&country=USA


mmoskal:
The AQI seems specific to the US (defined by EPA). EU has a different standard.  We could say that in the description that what is meant is the EPA one.

Also it's in principle possible for AQI to be fractional, so maybe use u16.16.

The EPA standard says you should average stuff over 1h, so not sure you'll get 1s update time.

See https://en.wikipedia.org/wiki/Air_quality_index#United_States

pelikhan:
I've added a min/max register to capture the various ranges defined for each countries. I don't think we want a different service for each range.

mmoskal:
Then maybe you should have a register that says which kind of measurement it is (and potentially change it). I don't think there's an easy conversion between different AQIs, but it seems easy to calculate different AQIs from the same PM2.5 etc data.

PR #1154, opened 2022-06-20
set design identifiers

mmoskal:
@pelikhan do you have some logic deriving id/filename from "name" field in JSON?


pelikhan:
Yes identifierToUrl or something

pelikhan (approved):

PR #1210, opened 2022-09-06
Service: Planar position

pelikhan:

### DIFF
--- a/services/planarposition.md
+++ b/services/planarposition.md
@@ -2,7 +2,8 @@
 
     identifier: 0x1dc37f55
     status: experimental
-    tags: input
+    extends: _sensor
+    group: movement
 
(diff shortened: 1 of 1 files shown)


services/planarposition.md
```diff
@@ -0,0 +1,19 @@
+# Planar position
+
+    identifier: 0x1dc37f55
+    status: experimental
+    tags: input
+
+A sensor that repors 2D position, typically an optical mouse tracking sensor.
+
+The sensor starts at an arbitrary origin (0,0) and reports the distance from that origin.
+
```
mmoskal, line 10: I wonder if we should suggest something like this:
```suggestion
The `streaming_interval` is respected when the position is changing. When the position is not changing, the streaming interval may be throttled to `500ms`.
```

mmoskal (approved):

PR #1231, opened 2022-12-19
adding semversion runtime version

pelikhan:
Adding a semver runtime version.


mmoskal (approved):
it doesn't fit the current format I use but will fix it on my side

mmoskal:
actually I reordered the fields so it fits 32 bit binary format

PR #1278, opened 2023-05-05
Device: Forward Education Moisture Sensor v1.0

brian-inksmith:


pelikhan:
The picture looks clipped, is it a bug of my tool. Do you have an unclipped version?

brian-inksmith:
Oh that’s odd.  The attached one should be better.  They’re all a little washed out.  We’ll get better ones before launch.

Ill work on the moisture capacitive variant and SVG early next week.

PR #1351, opened 2024-02-11
Device: M5Stack Cardputer v1.0

reggi:
I pulled this in manually from the form.
The Product Identifier came up as `0x32f390a2` for me on the site and converted to base 10 🤷 not sure if this is right.


reggi:
cc https://github.com/microsoft/devicescript/issues/679

@microsoft-github-policy-service agree company="Microsoft"
@microsoft-github-policy-service agree

reggi:
I'm a little confused as if this should be an entry for the M5Stamp or the M5Cardputer. The stamp is the chip, the cardputer is the device with the lcd and keyboard.

PR #1360, opened 2024-03-07
double use of Productidentifier

MKleinSB:
Sorry,  the Productidentifier was used double by me


MKleinSB:
Somehow Callicolor has not been added to the device catalog and ist not found n the device dashboard.

grafik

PR #1378, opened 2024-07-16
Arcade shields

tballmsft:


devices/icshop/gamebitexpansionboardv10.jpg
pelikhan, line 1: we need a better picture with the proper form factor or it'll look wanky in the web site.

PR #1404, opened 2025-04-30
Device: Forward Education Lights v1.0

brian-forwardedu:


ssande-fwd:
Product ID should be 827772839 based on observation that it is supposed to correspond to firmware ID.

mmoskal (approved):

PR #1405, opened 2025-04-30
Device: Forward Education Learn to Solder Kit v1.0

brian-forwardedu:


ssande-fwd:
Product ID should be 884365785 based on observation that it is supposed to correspond to firmware ID.

pelikhan:
> Product ID should be 884365785 based on observation that it is supposed to correspond to firmware ID.

can you push a fix?

mmoskal (approved):

